PR #68, opened 2019-09-05
fix: enable mount options

### LINKED ISSUE #25: add mountOptions support

Code is here: https://github.com/csi-driver/blobfuse-csi-driver/blob/a005ac785bb280d1fa9dee434e78d968a11b7ad0/pkg/blobfuse/nodeserver.go#L117-L118, could refer to https://github.com/Azure/kubernetes-volume-drivers/tree/master/flexvolume/blobfuse#about

**Additional context**

ZeroMagic:
**What type of PR is this?**
/kind bug

Each parameter in cmd should be separated with a blank. Otherwise, it won't work.

Fixes #25 

**Release note**:
```
fix: enable mount options
```

### DIFF
--- a/charts/latest/blobfuse-csi-driver/templates/csi-blobfuse-node.yaml
+++ b/charts/latest/blobfuse-csi-driver/templates/csi-blobfuse-node.yaml
@@ -94,6 +94,8 @@ spec:
               name: blobfuse-bin
             - mountPath: /mnt
               name: blobfuse-cache
+            - mountPath: /etc/fuse.conf
+              name: fuse-conf
       volumes:
         - hostPath:
             path: /var/lib/kubelet/plugins/blobfuse.csi.azure.com
@@ -120,4 +122,7 @@ spec:
         - hostPath:
             path: /mnt
           name: blobfuse-cache
-          
+        - hostPath:
+            path: /etc/fuse.conf
+            type: FileOrCreate
+          name: fuse-conf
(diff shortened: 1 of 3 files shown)


charts/latest/blobfuse-csi-driver/templates/csi-blobfuse-node.yaml
```diff
@@ -120,4 +122,7 @@ spec:
         - hostPath:
             path: /mnt
           name: blobfuse-cache
-          
\ No newline at end of file
+        - hostPath:
```
andyzhangx: is this required?

marcoboffi:
Sorry but I don't understand very well how it's possible define mountOption.
I found in your code that tmp_path is set by default to /mnt. But how can I set other parameters? With values file during driver installation with helm?

andyzhangx:
> Sorry but I don't understand very well how it's possible define mountOption.
> I found in your code that tmp_path is set by default to /mnt. But how can I set other parameters (in particular file-cache-timeout-in-seconds)? With values file during driver installation with helm?

check example here: https://github.com/kubernetes-sigs/blobfuse-csi-driver/blob/master/deploy/example/storageclass-blobfuse-csi-mountoptions.yaml

marcoboffi:
Great! But "-o uid=1000" or "allow_other" option are mandatory to use /mnt (Azure VM ephimeral disk) ?

andyzhangx:
> Great! But "-o uid=1000" or "allow_other" option are mandatory to use /mnt (Azure VM ephimeral disk) ?

@marcoboffi they are not relevant. do you have any other requirement on this?

PR #100, opened 2020-01-18
Set bigger limit

andyzhangx:
**What type of PR is this?**
/kind cleanup

This PR:
 - Set bigger limit
 - NodePublishVolume mount refactor

Fixes #

**Release note**:
```
none
```


charts/latest/blobfuse-csi-driver/templates/csi-blobfuse-controller.yaml
```diff
@@ -134,8 +134,8 @@ spec:
               name: msi
           resources:
             limits:
-              cpu: 200m
-              memory: 200Mi
+              cpu: 2
+              memory: 2
```
ZeroMagic: If the memory unit is omitted, it will be measured in bytes. I think 2 is not reasonable.

andyzhangx: it's 2Gi, typo

PR #165, opened 2020-06-09
doc: rename storage class to blob

andyzhangx:
**What type of PR is this?**
/kind documentation

doc: rename storage class to blob

Fixes #

**Release note**:
```
none
```

### DIFF
--- a/deploy/example/deployment.yaml
+++ b/deploy/example/deployment.yaml
@@ -9,7 +9,7 @@ spec:
   resources:
     requests:
       storage: 1Gi
-  storageClassName: blobfuse.csi.azure.com
+  storageClassName: blob
 ---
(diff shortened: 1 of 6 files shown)


deploy/example/deployment.yaml
```diff
@@ -9,7 +9,7 @@ spec:
   resources:
     requests:
       storage: 1Gi
-  storageClassName: blobfuse.csi.azure.com
+  storageClassName: blob
```
ZeroMagic, line 12: blobfuse-csi?

andyzhangx: there is no blob in-tree driver, also we are going to develop blob based NFS service, use `blob` should make more sense

PR #180, opened 2020-07-13
feat: switch dep to go module

### LINKED ISSUE #20: Switch glide to go modules for dependency management

could refer to https://github.com/kubernetes/cloud-provider-azure/pull/142

**Additional context**

ZeroMagic:
**What type of PR is this?**
/kind feature

Switch dep to go module.
Remove snapshot related code in e2e framework, because it is unnecessary now.
Need to use go-toml less than v1.8, otherwise there will be an error when parsing the toml file.

Fixes #20 

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

Refer to https://github.com/kubernetes-sigs/azurefile-csi-driver/pull/236

**Release note**:
```
none
```


pkg/util/util.go
```diff
@@ -66,3 +70,13 @@ func GetMountOptions(options []string) string {
 	}
 	return str
 }
+
+func MakeDir(pathname string) error {
```
andyzhangx, line 74: could you add a unit for this function @Sakuralbj

hack/update-gomod.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
```
andyzhangx, line 1: shall we always use this tool when update-gomod?

ZeroMagic: No. Just use it to upgrade Kubernetes.

pkg/util/util.go
Sakuralbj: ok

PR #272, opened 2020-11-11
chore: customization for Azure Stack Cloud

jiaxunsongucb:
**What type of PR is this?**
/kind feature

As Azure Stack and public Azure are using different storage account versions, we need to do some customizations for Azure Stack.
1. Azure Stack only supports Storage, so forcibly change the account kind to Storage (instead of StorageV2 or BlockBlobStorage).
2. Mount SSL certs.
3. Add an environment variable AZURE_ENVIRONMENT_FILEPATH, which is only used for Azure Stack: https://github.com/kubernetes-sigs/blob-csi-driver/blob/1ac16497e06b2b3601d2d0030b31e5450956e511/vendor/github.com/Azure/go-autorest/autorest/azure/environments.go#L220
This is not ideal, but the env variable must be passed before the ARM Client is created.
4. Make changes to the bash scripts under the `hack` folder so that `make verify` can work in a brand new Ubuntu testing VM in the Azure Stack test pipeline.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/charts/latest/blob-csi-driver/templates/csi-blob-controller.yaml
+++ b/charts/latest/blob-csi-driver/templates/csi-blob-controller.yaml
@@ -95,6 +95,8 @@ spec:
                   optional: true
             - name: CSI_ENDPOINT
               value: unix:///csi/csi.sock
+            - name: AZURE_ENVIRONMENT_FILEPATH  # For Azure Stack Cloud
+              value: /etc/kubernetes/azurestackcloud.json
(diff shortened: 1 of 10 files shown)


andyzhangx:
can you fix the unit test first, run `hack/verify-all.sh` in your dev env.

andyzhangx (approved):
/lgtm
/approve

PR #274, opened 2020-11-13
fix: only use Standard_LRS storageAccountType on Azure Stack

jiaxunsongucb:
**What type of PR is this?**
/kind bug

Convert storageAccountType to Standard_LRS for Azure Stack as it only supports this account type.
Add limitations in README.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/blob/controllerserver.go
+++ b/pkg/blob/controllerserver.go
@@ -112,6 +112,7 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	}
 	if IsAzureStackCloud(d.cloud) {
 		accountKind = string(storage.Storage)
+		storageAccountType = "Standard_LRS"
 	}
 
 	tags, err := azure.ConvertTagsToMap(customTags)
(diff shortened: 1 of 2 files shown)


pkg/blob/controllerserver.go
```diff
@@ -112,6 +112,7 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	}
 	if IsAzureStackCloud(d.cloud) {
 		accountKind = string(storage.Storage)
+		storageAccountType = "Standard_LRS"
```
andyzhangx, line 115: shall we report error here?

jiaxunsongucb: Can we give users errors when they use the wrong account type? I've mentioned this in the README that under the hood the account type will be converted to `Standard_LRS` anyways. I thought for Azure Stack scenarios, simplicity rather than complexity is better.

andyzhangx (approved):
/ok-to-test
/lgtm
/approve

PR #295, opened 2020-12-22
fix: only deploy test pods on Linux nodes

songjiaxun:
**What type of PR is this?**

/kind bug

In tests, deploy pods/deployments only on Linux nodes.
For Azure Stack tests, we will spin up a separate buildx builder container with cert path mounted.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


andyzhangx (approved):
/ok-to-test
/lgtm
/approve

Issue #284, opened 2020-11-24
Remove the dependency of azure.json in node server


ZeroMagic:
**Is your feature request related to a problem?/Why is this needed**

If the customers just use the node server to automatically complete the mount command, they don't need to provide the azure.json. But now, in our code, it is still required to provide the config.
After removing the config, the node server will also work on the non-AKS clusters in static provision.
**Describe the solution you'd like in detail**

**Describe alternatives you've considered**

**Additional context**

ZeroMagic:
- [x] Feature Code 

- [ ] Feature Doc

andyzhangx:
azure.json is not a dependency for both controller and node servers now.

PR #308, opened 2021-01-07
feat: remove dependency of azure.json in node server

ZeroMagic:
**What type of PR is this?**

/kind feature

Ref: https://github.com/kubernetes-sigs/blob-csi-driver/issues/284
Remove the dependency of azure.json in node server.

For controller server, the azure.json is necessary. The file can be stored at the specified path or a kubernetes secrets.
For node server, the azure.json is unnecessary. If the file is not provided, need to provide the account key in volume attributions or a kubernetes secrets. Otherwise, the node has no permission to get and mount the blob container. 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
feat: remove dependency of azure.json in node server
```

### DIFF
--- a/pkg/blob/azure.go
+++ b/pkg/blob/azure.go
@@ -46,21 +46,22 @@ func IsAzureStackCloud(cloud *azureprovider.Cloud) bool {
 	return strings.EqualFold(cloud.Config.Cloud, "AZURESTACKCLOUD")
 }
 
-// GetCloudProvider get Azure Cloud Provider
-func GetCloudProvider(kubeconfig string) (*azureprovider.Cloud, error) {
+// getCloudProvider get Azure Cloud Provider
+func getCloudProvider(kubeconfig string) (*azureprovider.Cloud, error) {
 	kubeClient, err := getKubeClient(kubeconfig)
 	if err != nil && !os.IsNotExist(err) && err != rest.ErrNotInCluster {
 		return nil, fmt.Errorf("failed to get KubeClient: %v", err)
 	}
 
 	az := &azureprovider.Cloud{}
+
 	if kubeClient != nil {
 		klog.V(2).Infof("reading cloud config from secret")
 		az.KubeClient = kubeClient
 		az.InitializeCloudFromSecret()
 	}
 
-	if az.TenantID == "" || az.SubscriptionID == "" {
+	if az.TenantID == "" || az.SubscriptionID == "" || az.ResourceGroup == "" {
 		klog.V(2).Infof("could not read cloud config from secret")
 		credFile, ok := os.LookupEnv(DefaultAzureCredentialFileEnv)
 		if ok {
@@ -86,6 +87,7 @@ func GetCloudProvider(kubeconfig string) (*azureprovider.Cloud, error) {
 	if kubeClient != nil {
 		az.KubeClient = kubeClient
 	}
+
 	return az, nil
 }
 
--- a/pkg/blob/azure.go
+++ b/pkg/blob/azure.go
@@ -47,7 +47,9 @@ func IsAzureStackCloud(cloud *azureprovider.Cloud) bool {
 }
 
 // getCloudProvider get Azure Cloud Provider
-func getCloudProvider(kubeconfig string) (*azureprovider.Cloud, error) {
+func getCloudProvider(kubeconfig, nodeID string) (*azureprovider.Cloud, error) {
+	isController := (nodeID == "")
+
 	kubeClient, err := getKubeClient(kubeconfig)
 	if err != nil && !os.IsNotExist(err) && err != rest.ErrNotInCluster {
 		return nil, fmt.Errorf("failed to get KubeClient: %v", err)
@@ -71,21 +73,37 @@ func getCloudProvider(kubeconfig string) (*azureprovider.Cloud, error) {
 			klog.V(2).Infof("use default %s env var: %v", DefaultAzureCredentialFileEnv, credFile)
 		}
 
-		f, err := os.Open(credFile)
+		var f *os.File
+		f, err = os.Open(credFile)
+		if f != nil {
+			defer f.Close()
+		}
 		if err != nil {
 			klog.Errorf("Failed to load config from file: %s", credFile)
-			return nil, fmt.Errorf("Failed to load config from file: %s, cloud not get azure cloud provider", credFile)
-		}
-		defer f.Close()
-
-		klog.V(2).Infof("read cloud config from file: %s successfully", credFile)
-		if az, err = azureprovider.NewCloudWithoutFeatureGates(f); err != nil {
-			return az, err
+			err = fmt.Errorf("Failed to load config from file: %s, cloud not get azure cloud provider", credFile)
+		} else {
+			klog.V(2).Infof("read cloud config from file: %s successfully", credFile)
+			az, err = azureprovider.NewCloudWithoutFeatureGates(f)
 		}
 	}
 
 	if kubeClient != nil {
 		az.KubeClient = kubeClient
+	}
+	if isController {
+		if err != nil {
+			return az, err
+		}
+		// Disable UseInstanceMetadata for controller to mitigate a timeout issue using IMDS
+		// https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/168
+		klog.Infoln("disable UseInstanceMetadata for controller")
+		az.Config.UseInstanceMetadata = false
+		klog.Infoln("Starting the controller server...")
+	} else {
+		if err != nil {
+			klog.V(2).Infof("there is no azure.json provided for node server, error: %v", err)
+		}
+		klog.V(2).Infof("Starting the node server, nodeID is(%s)", nodeID)
 	}
 
 	return az, nil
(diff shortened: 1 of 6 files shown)


pkg/blob/azure.go
```diff
@@ -39,28 +39,30 @@ import (
 var (
 	DefaultAzureCredentialFileEnv = "AZURE_CREDENTIAL_FILE"
 	DefaultCredFilePath           = "/etc/kubernetes/azure.json"
+	storageEndpointNameToSuffix   = map[string]string{
+		"publiccloud":       "core.windows.net",
```
andyzhangx, line 43: if it's only endpoint name, why not let user provide end point by themselves, by default it's always `core.windows.net`

andyzhangx: another way is provide a new parameter(`volumeAttributes.storageEndpoint`), that would be more flexible, even node has `azure.json`, if user provides `storageEndpoint` in storage class, should use this value

pkg/blob/blob.go
```diff
@@ -115,17 +116,39 @@ func (d *Driver) Run(endpoint, kubeconfig string, testBool bool) {
...
 		klog.Infoln("disable UseInstanceMetadata for controller")
 		d.cloud.Config.UseInstanceMetadata = false
+		klog.Infoln("Starting the controller server...")
+	} else {
+		cloud, err := getCloudProvider(kubeClient)
+		if err == nil {
+			d.cloud = cloud
+		} else {
+			klog.Infoln("there is no azure.json provided for node server")
+			if kubeClient != nil {
```
andyzhangx, line 142:
```
d.cloud = &azure.Cloud
if kubeClient != nil {
  d.cloud.KubeClient = kubeClient
}
```

pkg/blob/blob.go
```diff
@@ -115,17 +116,39 @@ func (d *Driver) Run(endpoint, kubeconfig string, testBool bool) {
...
 		// https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/168
 		klog.Infoln("disable UseInstanceMetadata for controller")
 		d.cloud.Config.UseInstanceMetadata = false
+		klog.Infoln("Starting the controller server...")
+	} else {
+		cloud, err := getCloudProvider(kubeClient)
+		if err == nil {
+			d.cloud = cloud
+		} else {
+			klog.Infoln("there is no azure.json provided for node server")
```
andyzhangx, line 141: `klog.V(2).Infof(...`

pkg/blob/blob.go
```diff
@@ -115,17 +116,39 @@ func (d *Driver) Run(endpoint, kubeconfig string, testBool bool) {
...
+			if kubeClient != nil {
+				d.cloud = &azure.Cloud{
+					KubeClient: kubeClient,
+				}
+			} else {
+				d.cloud = &azure.Cloud{}
+			}
+		}
+
+		klog.Infoln("Starting the node server...NodeID is ", d.NodeID)
```
andyzhangx, line 151: `klog.V(2).Infof(Starting the node server, nodeID(%s)`

pkg/blob/azure.go
```diff
@@ -46,21 +46,22 @@ func IsAzureStackCloud(cloud *azureprovider.Cloud) bool {
...
 	az := &azureprovider.Cloud{}
+
 	if kubeClient != nil {
 		klog.V(2).Infof("reading cloud config from secret")
 		az.KubeClient = kubeClient
 		az.InitializeCloudFromSecret()
 	}
 
-	if az.TenantID == "" || az.SubscriptionID == "" {
+	if az.TenantID == "" || az.SubscriptionID == "" || az.ResourceGroup == "" {
```
andyzhangx, line 64: why add a new condition `az.ResourceGroup == ""` ?

pkg/blob/nodeserver.go
```diff
@@ -151,15 +153,19 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 	}
 
 	if strings.TrimSpace(blobStorageEndPoint) == "" {
+		blobStorageEndPoint = "core.windows.net"
```
andyzhangx, line 156: it's called `storageEndpointSuffix`, let's use official term:
```
StorageEndpointSuffix:        "core.windows.net",
```

pkg/blob/blob.go
```diff
@@ -46,6 +46,7 @@ const (
 	volumeIDTemplate           = "%s#%s#%s"
 	secretNameTemplate         = "azure-storage-account-%s-secret"
 	serverNameField            = "server"
+	blobStorageEndPointField   = "blobstorageendpoint"
```
andyzhangx, line 49: `storageendpointsuffix`

pkg/blob/blob.go
```diff
@@ -115,17 +116,35 @@ func (d *Driver) Run(endpoint, kubeconfig string, testBool bool) {
...
 		d.cloud.Config.UseInstanceMetadata = false
+		klog.Infoln("Starting the controller server...")
+	} else {
+		cloud, err := getCloudProvider(kubeconfig)
+		if err == nil {
+			d.cloud = cloud
+		} else {
+			klog.V(2).Infof("there is no azure.json provided for node server")
+			d.cloud = &azure.Cloud{}
+			kubeClient, err := getKubeClient(kubeconfig)
```
andyzhangx, line 138: what about move all these logic into `getCloudProvider` func:
```
isController := (d.NodeID == "")
getCloudProvider(kubeconfig, isController)
```

That would be more clear since there is already `kubeClient, err := getKubeClient(kubeconfig)` in `getCloudProvider` func

pkg/blob/azure.go
ZeroMagic: In the previous tests, without the field in azure.json, there was an error. After adding it, the driver worked fine.

andyzhangx: which case failed? that means `TenantID` and `SubscriptionID` is not empty, while `ResourceGroup` is empty.

ZeroMagic: Not the case in our tests. It is in the non-AKS clusters. They used their own custom azure.json to run the driver.

pkg/blob/blob.go
ZeroMagic: done.

pkg/blob/nodeserver.go
ZeroMagic: done

pkg/blob/blob.go
```diff
@@ -115,18 +116,53 @@ func (d *Driver) Run(endpoint, kubeconfig string, testBool bool) {
...
 	}
-	d.cloud = cloud
 
-	if d.NodeID == "" {
-		// Disable UseInstanceMetadata for controller to mitigate a timeout issue using IMDS
-		// https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/168
-		klog.Infoln("disable UseInstanceMetadata for controller")
-		d.cloud.Config.UseInstanceMetadata = false
-	}
+	// if err != nil {
```
andyzhangx, line 124: remove comments

pkg/blob/azure.go
```diff
@@ -70,22 +73,37 @@ func GetCloudProvider(kubeconfig string) (*azureprovider.Cloud, error) {
...
+		f, err = os.Open(credFile)
 		if err != nil {
 			klog.Errorf("Failed to load config from file: %s", credFile)
-			return nil, fmt.Errorf("Failed to load config from file: %s, cloud not get azure cloud provider", credFile)
+			err = fmt.Errorf("Failed to load config from file: %s, cloud not get azure cloud provider", credFile)
+		} else {
+			klog.V(2).Infof("read cloud config from file: %s successfully", credFile)
+			az, err = azureprovider.NewCloudWithoutFeatureGates(f)
 		}
 		defer f.Close()
```
andyzhangx, line 85: move this to L78

pkg/blob/blob.go
ZeroMagic: done

pkg/blob/azure.go
ZeroMagic: done

andyzhangx (approved):
/lgtm
/approve

pkg/blob/nodeserver.go
```diff
@@ -150,16 +152,20 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
...
 	}
 
-	if strings.TrimSpace(blobStorageEndPoint) == "" {
+	if strings.TrimSpace(storageEndpointSuffix) == "" {
+		storageEndpointSuffix = "core.windows.net"
+	}
+
+	if strings.TrimSpace(serverAddress) == "" {
 		// server address is "accountname.blob.core.windows.net" by default
-		blobStorageEndPoint = fmt.Sprintf("%s.blob.%s", accountName, d.cloud.Environment.StorageEndpointSuffix)
```
andyzhangx, line 155: I think we need to keep `d.cloud.Environment.StorageEndpointSuffix` if `storageEndpointSuffix` is empty, that will not break azure stack test @songjiaxun

PR #325, opened 2021-01-21
fix: Azure Stack supports Premium_LRS storage account type

songjiaxun:
**What type of PR is this?**

/kind bug
Add logic for Azure Stack to support Premium_LRS storage account type.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/blob/controllerserver.go
```diff
@@ -114,7 +114,14 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
 	if IsAzureStackCloud(d.cloud) {
 		accountKind = string(storage.Storage)
-		storageAccountType = "Standard_LRS"
+		if strings.HasPrefix(strings.ToLower(storageAccountType), "standard") && storageAccountType != string(storage.StandardLRS) {
+			klog.Warningf("Use sku %s instead as Azure Stack does not support %s.", storage.StandardLRS, storageAccountType)
+			storageAccountType = "Standard_LRS"
+		}
+		if strings.HasPrefix(strings.ToLower(storageAccountType), "premium") && storageAccountType != string(storage.PremiumLRS) {
+			klog.Warningf("Use sku %s instead as Azure Stack does not support %s.", storage.PremiumLRS, storageAccountType)
+			storageAccountType = "Premium_LRS"
```
andyzhangx, line 123: convert saliently may be not a good option, maybe add a supported account list and if it's not in the list, prompt error

songjiaxun: I agree. But I remember we had a discussion about this before -- for simplicity, internally convert the storage account type to a supported type, and mention that in the documentation is good enough for now. Otherwise, we also need to change the test cases for Azure Stack, which outweighs the benefits.

pkg/blob/controllerserver.go
```diff
@@ -114,7 +114,9 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	}
 	if IsAzureStackCloud(d.cloud) {
 		accountKind = string(storage.Storage)
-		storageAccountType = "Standard_LRS"
+		if storageAccountType != string(storage.StandardLRS) && storageAccountType != string(storage.PremiumLRS) {
+			return nil, status.Errorf(codes.InvalidArgument, fmt.Sprintf("Invalid skuName value: %s, as Azure Stack only supports %s and %s Storage Account types.", storageAccountType, storage.PremiumLRS, storage.StandardLRS))
+		}
```
songjiaxun, line 119: @andyzhangx I took your advice to throw errors when unsupported sku is requested. Accordingly, I changed the test logic a bit. Please let me know your thoughts. If this is OK, I will make a similar change against Azure Disk CSI.

pkg/blob/controllerserver.go
```diff
@@ -114,7 +114,9 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	}
 	if IsAzureStackCloud(d.cloud) {
 		accountKind = string(storage.Storage)
-		storageAccountType = "Standard_LRS"
+		if storageAccountType != "" && storageAccountType != string(storage.StandardLRS) && storageAccountType != string(storage.PremiumLRS) {
+			return nil, status.Errorf(codes.InvalidArgument, fmt.Sprintf("Invalid skuName value: %s, as Azure Stack only supports %s and %s Storage Account types.", storageAccountType, storage.PremiumLRS, storage.StandardLRS))
+		}
```
songjiaxun, line 119: @andyzhangx  I took your advice to throw errors when unsupported sku is requested. Accordingly, I changed the test logic a bit. Please let me know your thoughts. If this is OK, I will make a similar change against Azure Disk CSI.

andyzhangx (approved):
/lgtm
/approve

PR #345, opened 2021-02-16
feat: add imagepullsecrets in helm chart

dza89:
Signed-off-by: daan <dhapotter@gmail.com>

Add imagepullsecrets to daemonset and deployment


dza89:
I don't know what version to make? 0.12.0 or 1.0.1? 
Also, it's quite weird to put tgzs in your repository.

dza89:
Alright, I'm not going to work on this MR anymore because the amount of work you have to do to add three lines is just too much. 
So either you guys can fix it up, otherwise close this mr

andyzhangx:
no worry, I will fix it up

dza89:
@andyzhangx 

Thanks, sorry for my rant.
Seems like alot of azure charts are based on old helm boilerplating, because this is default in the latest boilerplating.

charts/latest/blob-csi-driver/templates/csi-blob-node.yaml
```diff
@@ -13,6 +13,10 @@ spec:
 {{ include "blob.labels" . | indent 6 }}
         app: csi-blob-node
     spec:
+      {{- if .Values.imagePullSecrets }}
```
andyzhangx, line 16: only need to make change under charts/latest, and then run `helm package charts/latest/blob-csi-driver -d charts/latest/` to update tgz file, thanks.

dza89: Done(?)

charts/latest/blob-csi-driver/Chart.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: latest
 description: Azure Blob Storage CSI driver
 name: blob-csi-driver
-version: v1.0.1
```
andyzhangx: revert this, no need to change version here.

charts/latest/blob-csi-driver/values.yaml
```diff
@@ -20,6 +20,9 @@ image:
     tag: v1.1.0
     pullPolicy: IfNotPresent
 
+imagePullSecrets: []
```
andyzhangx, line 23: it's better write like this:
https://github.com/helm/charts/blob/master/stable/cert-manager/values.yaml#L5-L9

dza89: i don't know what i'm looking at? It looks the same to me?
Are you referring to the comments?

andyzhangx: yes, just comments, it's a little better

dza89: Done

andyzhangx (approved):
/ok-to-test

Issue #392, opened 2021-04-24
collect blobfuse-proxy logs in e2e testing


andyzhangx:
**Is your feature request related to a problem?/Why is this needed**

**Describe the solution you'd like in detail**

Currently we can get blobfuse-proxy logs on the agent node by:
```console
sudo journalctl -u blobfuse-proxy -l > blobfuse-proxy.log
```

**Describe alternatives you've considered**

**Additional context**

cc @boddumanohar

andyzhangx:
fixed

PR #400, opened 2021-04-28
tail blobfuse-proxy logs at the start of blobfuse-proxy daemonset

boddumanohar:
**What type of PR is this?**
tail blobfuse-proxy logs at the start of blobfuse-proxy daemonset

/kind cleanup

**What this PR does / why we need it**: #392 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


andyzhangx (commented):
/ok-to-test

deploy/blobfuse-proxy/blobfuse-proxy.yaml
```diff
@@ -54,8 +54,10 @@ spec:
             systemctl start blobfuse-proxy
             echo "removing the file: /tmp/blobfuse-proxy-v0.1.0.deb"
             rm /tmp/blobfuse-proxy-v0.1.0.deb
-            echo "sleeping for 19 years"
-            sleep 600000000s
+            echo "sleeping for 10 secs"
+            sleep 10s
```
andyzhangx, line 58: that means it can only get logs in the exact time of after 10s, is there better way to get all logs after all tests completed?

andyzhangx:
should filter out `AZURE_STORAGE_ACCESS_KEY` in blobfuse-proxy logging and better keep other fields in the logs:
```
received auth env [AZURE_STORAGE_ACCESS_KEY=xxx AZURE_STORAGE_ACCOUNT=fusecd9d3dacdfc94335852 
```

deploy/blobfuse-proxy/blobfuse-proxy.yaml
boddumanohar: it is getting all the logs. In the above case, the sleep is to wait for the blobfuse-proxy service to come up. Although it happens much quicker than that.

boddumanohar:
@andyzhangx  The .deb file present in the master is a little old. So I updated it with the latest one. Now we can observe we are not leaking secrets in the logs. 

also, the current method we are using to distribute the .deb file is a little tedious. Because if someone made changes to the blobfuse-proxy then they have to build the new .deb themselves and update the links in `deploy/blobfuse-proxy/blobfuse-proxy.yaml`. And there is no way to guarantee its if the .deb file is legit. Instead its better if we have some automation around this.

deploy/blobfuse-proxy/blobfuse-proxy.yaml
```diff
@@ -44,7 +44,7 @@ spec:
             apt-get update
             apt-get install -y blobfuse=1.3.6
             # download blobfuse-proxy .deb package
-            wget https://github.com/kubernetes-sigs/blob-csi-driver/raw/master/deploy/blobfuse-proxy/v0.1.0/blobfuse-proxy-v0.1.0.deb -O /tmp/blobfuse-proxy-v0.1.0.deb
+            wget https://github.com/boddumanohar/blob-csi-driver/raw/issues/392/deploy/blobfuse-proxy/v0.1.0/blobfuse-proxy-v0.1.0.deb -O /tmp/blobfuse-proxy-v0.1.0.deb
```
boddumanohar, line 47: instead of creating a separate PR to to change: `boddumanohar` --> `kubernetes-sigs` . Can I update it right now? So that when the PR is merged, it will point to the correct file  itself.

andyzhangx:
> @andyzhangx The .deb file present in the master is a little old. So I updated it with the latest one. Now we can observe we are not leaking secrets in the logs.
> 
> also, the current method we are using to distribute the .deb file is a little tedious. Because if someone made changes to the blobfuse-proxy then they have to build the new .deb themselves and update the links in `deploy/blobfuse-proxy/blobfuse-proxy.yaml`. And there is no way to guarantee its if the .deb file is legit. Instead its better if we have some automation around this.

@boddumanohar could add a verification script in `hack/` folder, it could build deb file automatically and do md5sum check with existing one, if there is diff, then verification failed?

deploy/blobfuse-proxy/blobfuse-proxy.yaml
andyzhangx: ok

boddumanohar:
@andyzhangx  I will propose the changes discussed above in a new PR.

boddumanohar:
@andyzhangx do you want me to propose the changes discussed above in this PR itself or a new PR?

andyzhangx:
> @andyzhangx do you want me to propose the changes discussed above in this PR itself or a new PR?

@boddumanohar new PR is ok

andyzhangx (approved):
/lgtm
/approve

PR #412, opened 2021-05-04
add a new helm parameter for livenessProbe health port

boddumanohar:
**What type of PR is this?**

/kind feature

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


charts/latest/blob-csi-driver/values.yaml
```diff
@@ -98,6 +98,8 @@ node:
   affinity: {}
   nodeSelector: {}
   tolerations: []
+  livenessProbe:
+    healthPort: 29653
```
andyzhangx, line 102: should be `29633`

andyzhangx (approved):
/lgtm
/approve
/ok-to-test

Issue #406, opened 2021-04-30
`ClusterRoles` and `ClusterRoleBindings` wrongly contain the `namespace` in the `metadata` section


Raffo:
**What happened**:

Some [ClusterRoleBindings](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/charts/latest/azuredisk-csi-driver/templates/rbac-csi-snapshot-controller.yaml#L62) and `ClusterRoles` contain a `namespace` in the metadata section. This is "invalid" as the resources are **not** namespaced as mentioned in the documentation https://kubernetes.io/docs/reference/access-authn-authz/rbac/#clusterrole-example.

**What you expected to happen**:

I would expect the YAML for `ClusterRoles` and `ClusterRoleBindings` to not contain the `metadata.namespace` field.

**How to reproduce it**:

Just a visual inspection of https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/charts/latest/azuredisk-csi-driver/templates/rbac-csi-snapshot-controller.yaml#L62 and other files.

**Anything else we need to know?**:

I would love to understand what is the process to address that. Should I change the manifests in latest? Would a fix be backported? Happy to help once I understand how this is supposed to work.

**Environment**:
- CSI Driver version: v1.1.1
- Kubernetes version (use `kubectl version`): 1.19
- OS (e.g. from /etc/os-release): default AKS node OS

This is the same problem as https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/819

PR #414, opened 2021-05-05
Fix wrongly configured `namespace` for `ClusterRole`

Raffo:
**What type of PR is this?**

/kind bug

https://github.com/kubernetes-sigs/blob-csi-driver/pull/407 did not fully fix https://github.com/kubernetes-sigs/blob-csi-driver/issues/406. This PR addresses the remaining one. 

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

Please let me know if the version change is correct. It looks like we have `1.2.0` in latest but not an associated release, so I was unsure what to do. 

**Release note**:
```
none
```


charts/latest/blob-csi-driver/Chart.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: latest
 description: Azure Blob Storage CSI driver
 name: blob-csi-driver
-version: v1.2.0
+version: v1.2.1
```
andyzhangx, line 5: pls keep `v1.2.0`, only need to change `rbac-csi-blob-node.yaml`

Raffo: Updated.

andyzhangx (approved):
/ok-to-test
/lgtm
/approve

PR #416, opened 2021-05-06
chore: use mcr alpine image in blobfuse-proxy

andyzhangx:
**What type of PR is this?**
/kind cleanup

chore: use mcr alpine image in blobfuse-proxy

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/deploy/blobfuse-proxy/blobfuse-proxy.yaml
+++ b/deploy/blobfuse-proxy/blobfuse-proxy.yaml
@@ -55,7 +55,7 @@ spec:
             sleep 3s
             # tail blobfuse proxy logs
             journalctl -u blobfuse-proxy -f
-        image: alpine
+        image: mcr.microsoft.com/oss/nginx/nginx:1.17.3-alpine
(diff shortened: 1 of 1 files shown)


deploy/blobfuse-proxy/blobfuse-proxy.yaml
```diff
@@ -47,15 +47,14 @@ spec:
             echo "Enabling blobfuse proxy systemctl service"
             systemctl daemon-reload
             systemctl enable blobfuse-proxy
-            systemctl stop blobfuse-proxy
```
andyzhangx, line 50: @boddumanohar FYI. This could ensure when blobfuse-proxy daemonsest restart, `blobfuse-proxy` service won't restart, then original mount would be still valid.

boddumanohar: Yeah. Nice edge case. Thanks for  pointing out. :)

PR #445, opened 2021-05-27
fix: only use pod.Namespace in inline volume support

andyzhangx:
**What type of PR is this?**
/kind bug

fix: only use pod.Namespace in inline volume support

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


feiskyer (approved):
/lgtm

PR #593, opened 2021-12-27
cleanup: remove cpu limits

andyzhangx:
**What type of PR is this?**
/kind cleanup

cleanup: remove cpu limits
set unreasonable cpu limits would affect user's workload calculation, this PR removes the unnecessary cpu limits.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
cleanup: remove cpu limits
```


feiskyer:
Is it possible to use [VPA](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler) for setting the correct limits?

andyzhangx:
> Is it possible to use [VPA](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler) for setting the correct limits?

@feiskyer there is transient cpu spike, should remove such cpu limits as suggested by Robbie Zhang. And some cpu limits are set in daemonset, it's not suitable to use VPA.

PR #614, opened 2022-02-17
Remove emeritus_reviewers from OWNERS file

srinivas-pokala:
Signed-off-by: Srinivas Pokala <Pokala.Srinivas@ibm.com>

What type of PR is this?

/kind documentation

Which issue(s) this PR fixes:
Fixes https://github.com/kubernetes/community/issues/6334

### DIFF
--- a/vendor/k8s.io/component-base/config/OWNERS
+++ b/vendor/k8s.io/component-base/config/OWNERS
@@ -7,10 +7,6 @@ approvers:
 - api-approvers
 reviewers:
 - api-reviewers
-emeritus_reviewers:
-- luxas
-- mtaufen
-- sttts
 labels:
 - kind/api-change
 - sig/api-machinery


vendor/k8s.io/component-base/config/OWNERS
```diff
@@ -7,10 +7,6 @@ approvers:
 - api-approvers
 reviewers:
 - api-reviewers
-emeritus_reviewers:
```
andyzhangx, line 10: this is a vendor change, I think you need to upgrade component-base version, and then `go mod vendor`, it's not right to modify vendor file directly

srinivas-pokala:
@spzala

vendor/k8s.io/component-base/config/OWNERS
spzala: @srinivas-pokala the file that you are modifying is a vendor file. The changes to such files should be done in the vendor lib/project and then consumed through updated version of the lib/project as Andy mentioned. In your case, I think the file is already modified here, https://github.com/kubernetes/component-base/commit/e0e7987a4c535d2e44a219f6435b7e5588a39dd6 This changes will be reflected when the k8s.io/component-base version is updated in this repo.

srinivas-pokala: @andyzhangx , discarding my PR changes  as it already modified here https://github.com/kubernetes/component-base/commit/e0e7987a4c535d2e44a219f6435b7e5588a39dd6 and will reflect in  k8s.io/component-base

PR #615, opened 2022-02-17
fix: nodepublishvolume idempotent issue and add a test case

fuxiuyin:
**What type of PR is this?**

/kind bug

NodePublishVolume need to be idempotent. Currently NodePublishVolume will mount a volume to the same mountpoint repeatedly when it's called with the same parameters multiply time. This may happen when CO encounter some failures. Such as, when NodePublishVolume is called but timeout, CO will call it with same parameters again.

run this can repo this issue.

```shell
sudo  go test -v -timeout 30s -run ^TestNodePublishVolumeIdempotentMount$ sigs.k8s.io/blob-csi-driver/pkg/blob
```

This issue mainly due to `IsLikelyNotMountPoint` can't handle bind mount. See [IsLikelyNotMountPoint doc](https://pkg.go.dev/k8s.io/utils/mount#Mounter.IsLikelyNotMountPoint). So, list all the mount points to confirm the mountpoint exists or not.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/blob/nodeserver.go
+++ b/pkg/blob/nodeserver.go
@@ -461,13 +461,13 @@ func (d *Driver) ensureMountPoint(target string) (bool, error) {
 		return !notMnt, err
 	}
 
-	target, err = filepath.Abs(target)
+	targetAbs, err := filepath.Abs(target)
 	if err != nil {
 		return !notMnt, err
 	}
 
 	for _, mountPoint := range mountList {
-		if mountPoint.Path == target {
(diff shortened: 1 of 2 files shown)


andyzhangx (commented):
/ok-to-test

andyzhangx (approved):
/lgtm
/approve

PR #627, opened 2022-03-13
[WIP]feat: support blobfuse2

andyzhangx:
**What type of PR is this?**
/kind feature

feat: support blobfuse2

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
feat: support blobfuse2
```


deploy/blobfuse2/config.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - attr_cache
+  - azstorage
+
+libfuse:
+  attribute-expiration-sec: 120
+  entry-expiration-sec: 120
+  negative-entry-expiration-sec: 240
+
+file_cache:
+  path: /tmp/blobfuse/
```
andyzhangx, line 19: in blobfuse, we have `--tmp-path` mount option, do we have corresponding mount option in blobfuse2?

andyzhangx: let me remove this `path` parameter first.

deploy/blobfuse2/config.yaml
```diff
@@ -0,0 +1,23 @@
...
+  - file_cache
+  - attr_cache
+  - azstorage
+
+libfuse:
+  attribute-expiration-sec: 120
+  entry-expiration-sec: 120
+  negative-entry-expiration-sec: 240
+
+file_cache:
```
andyzhangx, line 18: there was `path` parameter under `file_cache`, I have removed it since it could be replaced by `--tmp-path` in blobfuse2 mount option.

deploy/blobfuse2/config.yaml
```diff
@@ -0,0 +1,23 @@
+allow-other: true
+
+logging:
+  type: base
```
vibhansa-msft, line 4: Keeping default logging policy as "base" means logs will be dumped into a file, we shall rather go for "syslog" as default logging policy.

deploy/blobfuse2/config.yaml
```diff
@@ -0,0 +1,23 @@
...
+  level: log_debug
```
vibhansa-msft, line 5: Do we want to keep log_debug as default option, it will dump good amount of logs and overall performance will also hit. What do we keep as default logging for blobfuse ? if nothing is configured then its default to log_warning.

deploy/blobfuse2/config.yaml
```diff
@@ -0,0 +1,23 @@
...
+  - azstorage
+
+libfuse:
+  attribute-expiration-sec: 120
+  entry-expiration-sec: 120
+  negative-entry-expiration-sec: 240
+
+file_cache:
+  timeout-sec: 120
+  max-size-mb: 4096
```
vibhansa-msft, line 20: By putting max-size-mb limit we are saying temp cache of blobfuse is limited to 4GB only. By default in blobfusev1 there is no such limits unless user explicitly wants to contain the size. For symmetry with v1 we can just remove this config from here.

deploy/blobfuse2/config.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+attr_cache:
+  timeout-sec: 7200
```
vibhansa-msft, line 23: Do we need to give azstorage config where user gives storage account details, or those have been ignored explicitly.

pkg/blob/blob.go
```diff
@@ -118,7 +120,7 @@ const (
 )
 
 var (
-	supportedProtocolList = []string{fuse, nfs}
+	supportedProtocolList = []string{fuse, fuse2, nfs}
```
vibhansa-msft, line 123: I am not sure about one thing here. Blobfuse (v1) needs "fuse (libfuse)" to be installed on the system. Is that done automatically by CSI driver or user does it manually, If its done automatically then we need to install "fuse3" on the system as well for v2 to work. On systems where "fuse3" is not supported, v2 can work with fuse2 however for that we have a seperate binary.

pkg/blobplugin/Dockerfile
```diff
@@ -34,7 +34,9 @@ ARG ARCH=amd64
 RUN if [ "$ARCH" = "amd64" ] ; then \
   clean-install libcurl4-gnutls-dev && \
   wget -O /blobfuse-proxy/packages-microsoft-prod.deb https://packages.microsoft.com/config/ubuntu/18.04/packages-microsoft-prod.deb && \
-  dpkg -i /blobfuse-proxy/packages-microsoft-prod.deb && apt update && apt install blobfuse fuse -y && apt remove wget -y; fi
+  dpkg -i /blobfuse-proxy/packages-microsoft-prod.deb && apt update && apt install blobfuse blobfuse2 fuse -y && apt remove wget -y; fi
```
vibhansa-msft, line 37: blobfuse blobfuse2 fuse "fuse3" as well

deploy/blobfuse2/config.yaml
andyzhangx: I think that's the main reason I use this config.yaml for blobfuse2, let me try remove this config file and check whether blobfuse2 e2e test could pass: https://github.com/Azure/azure-storage-fuse/issues/734#issuecomment-1066061229

pkg/blobplugin/Dockerfile
andyzhangx: does `fuse3` package exist on ubuntu 18.04? I could not find it.
```
# apt update
Hit:1 http://azure.archive.ubuntu.com/ubuntu bionic InRelease
Get:2 http://azure.archive.ubuntu.com/ubuntu bionic-updates InRelease [88.7 kB]
Get:3 http://azure.archive.ubuntu.com/ubuntu bionic-backports InRelease [74.6 kB]
Hit:4 https://packages.microsoft.com/repos/azure-cli bionic InRelease
Hit:5 https://packages.microsoft.com/ubuntu/18.04/prod bionic InRelease
Get:6 https://aquasecurity.github.io/trivy-repo/deb bionic InRelease [3,054 B]
Hit:7 https://cli.github.com/packages bionic InRelease
Hit:8 https://download.docker.com/linux/ubuntu focal InRelease
Get:9 http://security.ubuntu.com/ubuntu bionic-security InRelease [88.7 kB]
Get:10 https://baltocdn.com/helm/stable/debian all InRelease [7,652 B]
Get:11 https://baltocdn.com/helm/stable/debian all/main amd64 Packages [2,872 B]
Fetched 266 kB in 1s (285 kB/s)
Reading package lists... Done
Building dependency tree
Reading state information... Done
89 packages can be upgraded. Run 'apt list --upgradable' to see them.
# apt search fuse3
Sorting... Done
Full Text Search... Done
# apt install fuse3
Reading package lists... Done
Building dependency tree
Reading state information... Done
E: Unable to locate package fuse3
# uname -a
Linux andydev 5.4.0-1077-azure #80~18.04.1-Ubuntu SMP Wed Apr 13 02:07:09 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

vibhansa-msft: No for Ubn-18 fuse3 is not available, but for Ubn20/22 it is. When you take blobfuse2 binary from ubn18 path it will be fuse2 compatible. just make sure we dont mix and match like taking blobfuse2 binary from ubn20 and packaging it for ubn18.

deploy/blobfuse2/config.yaml
gapra-msft: We support all the env variables that v1 supports so if that is the scenario for auth the customer uses with CSI driver it should be fine.

pkg/blobplugin/Dockerfile
andyzhangx: thanks for the clarification, that means Ubuntu OS takes those dependencies, CSI driver won't need to care about the dependencies. We will install fues3 when Ubuntu 22.04 is supported on AKS.

pkg/blob/blob.go
andyzhangx: Ubuntu OS manages those dependencies

PR #691, opened 2022-06-09
feat: combine blobfuse-proxy daemonset with driver node daemonset

### LINKED ISSUE #692: No need to use daemonset to install and start blobfuse-proxy

Add an initContainer in blob-csi-node to install and start blobfuse-proxy service on host.

**Additional context**

cvvz:
**What type of PR is this?**

/kind feature

No need a standalone daemonset for blobfuse-proxy. Change blobfuse-proxy from daemonset to initContainer.

Fixes #692 

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Other changes includes:**
1. Change the way to install blobfuse-proxy from using deb package manager to binary installation.

**Release note**:
```
Change blobfuse-proxy from daemonset to initContainer
```


andyzhangx:
```
Error: INSTALLATION FAILED: YAML parse error on blob-csi-driver/templates/csi-blob-node.yaml: error converting YAML to JSON: yaml: line 54: did not find expected '-' indicator
```

build/blobplugin/Dockerfile
```diff
@@ -18,10 +18,13 @@ ARG ARCH=amd64
 ARG binary=./_output/${ARCH}/blobplugin
 COPY ${binary} /blobplugin
 
+COPY ./build/blobplugin/blobfuse-proxy /blobfuse-proxy
```
andyzhangx, line 21: would you keep the original path: `pkg/blobplugin`? that's what other CSI drivers are like.

pkg/blobfuse-proxy/debpackage/DEBIAN/control
```diff
@@ -1,7 +0,0 @@
-Package: blobfuse-proxy
```
andyzhangx, line 1: this is not needed any more?

charts/latest/blob-csi-driver/templates/csi-blob-node.yaml
```diff
@@ -188,6 +225,17 @@ spec:
             {{- end }}
           resources: {{- toYaml .Values.node.resources.blob | nindent 12 }}
       volumes:
+{{- if .Values.node.enableBlobfuseProxy }}
+        - name: host-tmp
+          hostPath:
+            path: /tmp/
```
andyzhangx, line 231: the sidecar container also has /tmp dir, is it necessary to mount /tmp/ on agent node?

andyzhangx (commented):
latest chart config has changed, pls run "helm package charts/latest/blob-csi-driver -d charts/latest/" to update tgz file

charts/latest/blob-csi-driver/values.yaml
```diff
@@ -147,8 +147,6 @@ node:
   nodeSelector: {}
   tolerations:
     - operator: "Exists"
-  livenessProbe:
```
andyzhangx, line 150: why remove such port, it's used here:
https://github.com/kubernetes-sigs/blob-csi-driver/blob/2832df4ef82364d967c09d84790690912c573dfd/charts/latest/blob-csi-driver/templates/csi-blob-node.yaml#L129-L132

build/blobplugin/Dockerfile
cvvz: I move the dockerfile and package configurations to `build` directory according to the suggestion from  https://github.com/golang-standards/project-layout, looks this can make the project layout more clear.

pkg/blobfuse-proxy/debpackage/DEBIAN/control
cvvz: Yep, I install and start blobfuse-proxy by using binary file directly.

charts/latest/blob-csi-driver/templates/csi-blob-node.yaml
cvvz: This can be any Path in the host, since it is used to put blobfuse `deb` file and install it to the host. It is to keep the same as original [blobfuse-proxy](https://github.com/kubernetes-sigs/blob-csi-driver/blob/master/deploy/blobfuse-proxy.yaml#L112-L115).

charts/latest/blob-csi-driver/values.yaml
cvvz: VSCode hints that it duplicates with https://github.com/kubernetes-sigs/blob-csi-driver/blob/master/charts/latest/blob-csi-driver/values.yaml#L114-L115

andyzhangx: yes, that's correct

charts/latest/blob-csi-driver/templates/csi-blob-node.yaml
andyzhangx: then let's remove `/tmp`, it should work inside sidecar container

cvvz: We still need a hostPath to put blobfuse deb file in, a path only exists in sidecar container may not work.

build/blobplugin/Dockerfile
andyzhangx: o, I just realized that our MCR image build depends on this Dockerfile path, it's better not change the path, otherwise MCR image build should have two Dockerfile paths for new and old path versions, let's keep as it.

deploy/csi-blob-node.yaml
```diff
@@ -34,6 +35,32 @@ spec:
...
+      initContainers:
+        - name: install-blobfuse-proxy
+          image: mcr.microsoft.com/k8s/csi/blob-csi:latest
+          imagePullPolicy: IfNotPresent
+          command:
+            - "/blobfuse-proxy/init.sh"
+          securityContext:
+            privileged: true
+          env:
+            - name: DEBIAN_FRONTEND
```
andyzhangx, line 47: It's better provide an env or some switch to disable blobfuse-proxy in deploy/*.yaml config, similar to charts config, since this blobfuse-proxy script actually does not work on non-Ubuntu node, wdyt?

build/blobplugin/Dockerfile
cvvz: OK, that sounds reasonable.

deploy/csi-blob-node.yaml
cvvz: How about tracking it in another issue? And I will fix it after this PR merged.

andyzhangx: sure, pls file an issue, thanks.

andyzhangx:
latest chart config has changed, pls run "helm package charts/latest/blob-csi-driver -d charts/latest/" to update tgz file

deploy/csi-blob-node.yaml
cvvz: https://github.com/kubernetes-sigs/blob-csi-driver/issues/695

cvvz:
tgz file updated

andyzhangx (approved):
/lgtm
/approve

Issue #693, opened 2022-06-10
blobfuse-proxy restart will make blobfuse exit


cvvz:
**What happened**:
blobfuse-proxy restart will cause blobfuse exit, and influence existing mount.

**What you expected to happen**:
blobfuse-proxy restart should not influence existing mount.

**How to reproduce it**:
1. Install blob csi driver with blobfuse-proxy enabled.
2. Create a Pod mount blob PVC using blobfuse protocol.
3. Enter the host.
4. Kill or restart blobfuse-proxy process.

**Anything else we need to know?**:

**Environment**:
- CSI Driver version:
latest
- Kubernetes version (use `kubectl version`):
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #707, opened 2022-06-25
fix: open file num limit setting issue and KillMode in blobfuse-proxy

### LINKED ISSUE #706: AKS | EMFILE: too many open files -> Blobfuse proxy making troubles if nodejs application is opening a few thousand files
**What happened**:
If we deploy the blob-csi-driver with enabled blobfuseproxy we are seeing the error `EMFILE: too many open files` after opening a few thousand files in our deployed nodejs application:
`helm install blob-csi-driver blob-csi-driver/blob-csi-driver --namespace kube-system --version v1.15.0 --set node.enableBlobfuseProxy=true`

If we deploy the blob-csi-driver without enabling the blobfuseproxy we do not have the problem:
`helm install blob-csi-driver blob-csi-driver/blob-csi-driver --namespace kube-system --version v1.15.0`

We expect to be able to open more files also when blobfuseproxy is enabled

**How to reproduce it**:

- Spin up an AKS cluster
- Install blob-csi-driver with enabled blobfuseproxy: `helm install blob-csi-driver blob-csi-driver/blob-csi-driver --namespace kube-system --version v1.15.0 --set node.enableBlobfuseProxy=true`
- Deploy an application that is opening a few thousand files inside a mounted blob storage

**Environment**:

- CSI Driver version: v1.15.0
- Kubernetes version (use `kubectl version`): v1.22.6 (AKS)
- OS (e.g. from /etc/os-release): Ubuntu 18.04.6 LTS
- Kernel (e.g. `uname -a`): Linux 5.4.0-1083-azure

- We already tried to increase the maxOpenFileNum settings of the helm-chart but this does not help:
  `helm install blob-csi-driver blob-csi-driver/blob-csi-driver --namespace kube-system --version v1.15.0 --set node.enableBlobfuseProxy=true --set node.blobfuseProxy.setMaxOpenFileNum=true --set node.blobfuseProxy.maxOpenFileNum=999000000`
- The problem seem to be similar to this closed (but unresolved) issue: https://github.com/Azure/azure-storage-fuse/issues/653
- As the problem does only happen if the blobfuseproxy is enalbed I guess the problem is coming from the blobfuseproxy
- When the error occurs `lsof | wc -l` is only counting 69136 elements which is far below the maxOpenFileNum.

**Possible reason:**
So far I could see, the blobfuse-proxy is installed as an systemd-service on the nodes. I am not sure if everybody is aware that processes started by systemd are limited regarding open-files from the systemd-daemon itself, see https://manpages.ubuntu.com/manpages/bionic/man5/systemd.exec.5.html#process%20properties

As far as I could see the current blobfuse-proxy service is described like this:
```
[Unit]
Description=Blobfuse proxy service

[Service]
ExecStart=/usr/bin/blobfuse-proxy --v=5 --blobfuse-proxy-endpoint=unix://var/lib/kubelet/plugins/blob.csi.azure.com/blobfuse-proxy.sock

[Install]
WantedBy=multi-user.target
```

That means systemd-limits are not explicitly set and are defaulting maybe to a low value. Setting `sysctl -w fs.file-max=9000000` which the driver init is doing will not help to get around this systemd-limitation. The containerd-service on AKS nodes is e.g. deployed like this:
```
[Unit]
Description=containerd daemon
After=network.target

[Service]
ExecStartPre=/sbin/modprobe overlay
ExecStart=/usr/bin/containerd
Delegate=yes
KillMode=process
Restart=always
OOMScoreAdjust=-999
# Having non-zero Limit*s causes performance problems due to accounting overhead
# in the kernel. We recommend using cgroups to do container-local accounting.
LimitNPROC=infinity
LimitCORE=infinity
LimitNOFILE=infinity
TasksMax=infinity

[Install]
WantedBy=multi-user.target
```

As one can see the LimitNOFILE and also other configuration values are explicitly set to infinity which allows the deployments (running in the same context) to open more files than the systemd default values.

I could not test but maybe somewhere in this systemd limitations the problem can be found. Let me know if I can give further useful information.

andyzhangx:
**What type of PR is this?**
/kind bug

fix: open file num limit setting issue and KillMode in blobfuse-proxy

Fixes #706, #693

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

<details>

https://manpages.ubuntu.com/manpages/jammy/man5/systemd.kill.5.html
https://www.freedesktop.org/software/systemd/man/systemd.exec.html

```
KillMode=
           Specifies how processes of this unit shall be killed. One of control-group, mixed,
           process, none.

           If set to control-group, all remaining processes in the control group of this unit
           will be killed on unit stop (for services: after the stop command is executed, as
           configured with ExecStop=). If set to mixed, the SIGTERM signal (see below) is sent to
           the main process while the subsequent SIGKILL signal (see below) is sent to all
           remaining processes of the unit's control group. If set to process, only the main
           process itself is killed (not recommended!). If set to none, no process is killed
           (strongly recommended against!). In this case, only the stop command will be executed
           on unit stop, but no process will be killed otherwise. Processes remaining alive after
           stop are left in their control group and the control group continues to exist after
           stop unless empty.
```
</details>

**Release note**:
```
fix: open file num limit setting issue and KillMode in blobfuse-proxy
```

### DIFF
--- a/pkg/blobfuse-proxy/blobfuse-proxy.service
+++ b/pkg/blobfuse-proxy/blobfuse-proxy.service
@@ -3,6 +3,16 @@ Description=Blobfuse proxy service
 
 [Service]
 ExecStart=/usr/bin/blobfuse-proxy --v=5 --blobfuse-proxy-endpoint=unix://var/lib/kubelet/plugins/blob.csi.azure.com/blobfuse-proxy.sock
+Delegate=yes
+KillMode=process
+Restart=always
+OOMScoreAdjust=-999
+# Having non-zero Limit*s causes performance problems due to accounting overhead
+# in the kernel. We recommend using cgroups to do container-local accounting.
+LimitNPROC=infinity
+LimitCORE=infinity
+LimitNOFILE=infinity
+TasksMax=infinity
 
 [Install]
 WantedBy=multi-user.target


Babbel404:
Hello @andyzhangx, do you have an estimate or an update for when this change will be published in a new version (potentially the release of v1.15.0)

I have been doing some debugging for failed mounts on my cluster, and I have determined the issue to be ultimately caused by this file limit bug.

andyzhangx:
@Babbel404 we will publish v1.15.0 release this week.

PR #722, opened 2022-07-25
e2e test for sas token

cvvz:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/blob/blob.go
```diff
@@ -317,7 +317,7 @@ func checkContainerNameBeginAndEnd(containerName string) bool {
 
 // isSASToken checks if the key contains the patterns. Because a SAS Token must have these strings, use them to judge.
 func isSASToken(key string) bool {
-	return strings.Contains(key, "?sv=")
+	return strings.HasPrefix(key, "?")
```
andyzhangx, line 320: is the previous test failure due to this issue?

andyzhangx, line 320: add a link to sas token format: https://docs.microsoft.com/en-us/rest/api/eventhub/generate-sas-token
I think it's better set like this: `strings.Contains(key, "&sig=")` per the above sas token format doc

cvvz: It's one of the reason. The other one is my mistake.

cvvz: The SAS token is actually a query string in URL, the format is like: `?sv={}&ss={}&srt={}&sp={}&se={}&st={}&spr={}&sig={}`. I guess maybe `sig=` could be put right after question mark `?`. Also I think question mark won't appear in access token, so I make the change in this way.

cvvz: Also I will add a link to SAS token format in the comment.

andyzhangx: the real sas token format actually only contains a few of the above const format, we may only check a few, e.g. `se=`, `sig=`,  and question mark check may be not necessary, as long as the former 2 fixed strings are there, then it's a sas token

andyzhangx: nvm, let's just use prefix `?` check to make it simple.

test/e2e/testsuites/pre_provisioned_sastoken_tester.go
```diff
@@ -0,0 +1,107 @@
...
+	sasURL, err := serviceClient.GetSASURL(
+		azblob.AccountSASResourceTypes{Object: true, Service: true, Container: true},
+		azblob.AccountSASPermissions{Read: true, List: true, Write: true, Delete: true, Add: true, Create: true, Update: true},
+		time.Now(), time.Now().Add(10*time.Hour))
+	framework.ExpectNoError(err)
+	ginkgo.By("SAS URL: " + sasURL)
+	u, err := url.Parse(sasURL)
+	framework.ExpectNoError(err)
+	sasToken := "?" + u.RawQuery
+	ginkgo.By("SAS Token: " + sasToken)
```
andyzhangx, line 105: remove any secret and sas token print log

andyzhangx (approved):
/lgtm
/approve

PR #728, opened 2022-08-04
Added backtick on errors.Wrap

### LINKED ISSUE #727: Not used backtick on errors.Wrap function in README.md

Screenshot (254)

file :- https://github.com/kubernetes-sigs/blob-csi-driver/tree/master/vendor/github.com/pkg/errors

It would be better if we use `errors.Wrap` instead of without backtick errors.Wrap

bishal7679:
**What type of PR is this?**
/kind documentation

backtick added on all errors.Wrap

## Fixes #727 

**Requirements**:
  
- [x] includes documentation

### DIFF
--- a/vendor/github.com/pkg/errors/README.md
+++ b/vendor/github.com/pkg/errors/README.md
@@ -14,7 +14,7 @@ which applied recursively up the call stack results in error reports without con
 
 ## Adding context to an error
 
-The errors.Wrap function returns a new error that adds context to the original error. For example
+The `errors.Wrap` function returns a new error that adds context to the original error. For example
 ```go
 _, err := ioutil.ReadAll(r)
 if err != nil {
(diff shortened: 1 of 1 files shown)


vendor/github.com/pkg/errors/README.md
```diff
@@ -14,7 +14,7 @@ which applied recursively up the call stack results in error reports without con
 
 ## Adding context to an error
 
-The errors.Wrap function returns a new error that adds context to the original error. For example
+The `errors.Wrap` function returns a new error that adds context to the original error. For example
```
andyzhangx, line 17: pls do not change vendor file directly, if there is issue, pls change dependent project first.

bishal7679: Okay I will follow this

PR #729, opened 2022-08-06
feat: support cross subscription with management API

cvvz:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/blob/blob.go
```diff
@@ -266,20 +266,20 @@ func (d *Driver) Run(endpoint, kubeconfig string, testBool bool) {
 }
 
 // GetContainerInfo get container info according to volume id, e.g.
-// input: "rg#f5713de20cde511e8ba4900#containerName#uuid"
-// output: rg, f5713de20cde511e8ba4900, containerName
-// input: "rg#f5713de20cde511e8ba4900#containerName#uuid#namespace"
-// output: rg, f5713de20cde511e8ba4900, containerName, namespace
-func GetContainerInfo(id string) (string, string, string, string, error) {
+// input: "subsID#rg#f5713de20cde511e8ba4900#containerName#uuid"
```
andyzhangx, line 269: add subsID to the beginning of the volumeID would be a breaking change, suggest appending subsID to the original volumeID format, and add ut test to check original format works 

also if subsID is same as cluster, we should keep the same format, it's not necessary to always append subsID to the volumeID since in most cases, it's in same subs.

pkg/blob/controllerserver.go
```diff
@@ -163,6 +163,8 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		if !storeAccountKey {
 			return nil, status.Errorf(codes.InvalidArgument, fmt.Sprintf("storeAccountKey must set as true in cross subscription(%s)", subsID))
 		}
+	} else {
+		subsID = d.cloud.SubscriptionID
```
andyzhangx, line 167: it should be ok set `subsID` as empty which mean this driver would use default subsID as current cluster, blob client supports `subsID` as empty, right?

andyzhangx:
btw, also add a volumeID format doc here: https://github.com/kubernetes-sigs/blob-csi-driver/blob/master/docs/driver-parameters.md

pkg/blob/blob.go
cvvz: Good point. How about I use $ instead of #, that is to say add `subsID$` to the beginning of the volumeID? That should be backward compatible.

pkg/blob/controllerserver.go
cvvz: That's right. I will leave it empty.

docs/driver-parameters.md
```diff
@@ -51,6 +51,8 @@ k8s-azure-created-by: azure
 pvc-92a4d7f2-f23b-4904-bad4-2cbfcff6e388
 ```
 
+ - VolumeId is identifier for the volume generated after volume is created by plugin successfully, the format of VolumeId is: `rg#accountName#containerName#uuid#secretNamespace#subsID`
```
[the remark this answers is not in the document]
andyzhangx: and if current account is in the same subs as the driver, `#subsID` would be empty.

pkg/blob/blob.go
```diff
@@ -266,20 +266,25 @@ func (d *Driver) Run(endpoint, kubeconfig string, testBool bool) {
...
-// input: "rg#f5713de20cde511e8ba4900#containerName#uuid#namespace"
-// output: rg, f5713de20cde511e8ba4900, containerName, namespace
-func GetContainerInfo(id string) (string, string, string, string, error) {
+// input: "rg#f5713de20cde511e8ba4900#containerName#uuid#"
+// output: rg, f5713de20cde511e8ba4900, containerName, "" , ""
+// input: "rg#f5713de20cde511e8ba4900#containerName#uuid#namespace#"
+// output: rg, f5713de20cde511e8ba4900, containerName, namespace, ""
+// input: "rg#f5713de20cde511e8ba4900#containerName#uuid#namespace#subsID"
+// output: rg, f5713de20cde511e8ba4900, containerName, namespace, subsID
+func GetContainerInfo(id string) (string, string, string, string, string, error) {
```
andyzhangx, line 275: add comments here:
the format of VolumeId is: `rg#accountName#containerName#uuid#secretNamespace#subsID`

andyzhangx (commented):
need to add some comments of the volumeid format, other parts LGTM

docs/driver-parameters.md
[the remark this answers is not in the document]
andyzhangx: btw, we should add same format notes in azure file driver too.

andyzhangx:
also change the PR title to `feat: support cross subscription with management API`

docs/driver-parameters.md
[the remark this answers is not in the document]
cvvz: Sure, I will do that for azure file driver.

pkg/blob/controllerserver.go
```diff
@@ -442,7 +442,8 @@ func (d *Driver) ValidateVolumeCapabilities(ctx context.Context, req *csi.Valida
 		if resourceGroupName == "" {
 			resourceGroupName = d.cloud.ResourceGroup
 		}
-		blobContainer, err := d.cloud.BlobClient.GetContainer(ctx, resourceGroupName, accountName, containerName)
+		blobContainer, retryErr := d.cloud.BlobClient.GetContainer(ctx, subsID, resourceGroupName, accountName, containerName)
+		err = retryErr.Error()
```
andyzhangx, line 446: `retryErr` could be nil, there retryErr.Error() would panic

cvvz: this scenario is handled in [Error()](https://github.com/kubernetes-sigs/blob-csi-driver/blob/master/vendor/sigs.k8s.io/cloud-provider-azure/pkg/retry/azure_error.go#L77-L79) , that is when retryErr is nil, retryErr.Error() will be nil too.

andyzhangx (approved):
/lgtm
/approve

PR #723, opened 2022-07-25
WIP Improved Unit-Test Coverage for Issue #703 (please see #736)

aarongalang:
**What type of PR is this?**
test

**What this PR does / why we need it**: 
Adds unit-tests to pkg folder for improved coverage

**Which issue(s) this PR fixes**: 
#703 

Fixes #

**Requirements**:
  
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Special notes for your reviewer**:

**Release note**:
```
none
```

Issue #703, opened 2022-06-17
improve unit test coverage


andyzhangx:
**Is your feature request related to a problem?/Why is this needed**

**Describe the solution you'd like in detail**

It's now 69%, target num: 85%

https://coveralls.io/github/kubernetes-sigs/blob-csi-driver?branch=master
![image](https://user-images.githubusercontent.com/4178417/174219842-32fb9e7d-3abf-4c20-a0ad-15f7e2f32b07.png)

**Describe alternatives you've considered**

**Additional context**

PR #736, opened 2022-08-19
Improved Unit-Test Coverage for Issue #703

aarongalang:
**What type of PR is this?**
/kind test

Adds unit-tests to pkg folder for improved coverage

#703 

Fixes #

**Requirements**:
  
- [ ] adds unit tests

This PR had to be created because my previous pr had a commit that used my local computer as the author, rather than my git user id. As a result, only that one commit did not have my cla, causing the pr to be mergeable
please see pr #723 for actual commit messages
**Release note**:
```
none
```


aarongalang:
accidentally changed the non-test files, undoing them now

pkg/blob/blob.go
```diff
@@ -1,12 +1,9 @@
 /*
 Copyright 2019 The Kubernetes Authors.
-
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
-
```
mukhoakash, line 7: revert the changes in this file

mukhoakash: Are these created by lint?

pkg/blob/controllerserver.go
```diff
@@ -1,12 +1,9 @@
 /*
 Copyright 2017 The Kubernetes Authors.
-
```
mukhoakash, line 3: Revert, unless created by lint fixes.

pkg/blob/nodeserver.go
```diff
@@ -1,12 +1,9 @@
 /*
 Copyright 2017 The Kubernetes Authors.
-
```
mukhoakash, line 3: Same. Revert unless created by lint fix.

test/utils/credentials/credentials.go
```diff
@@ -1,12 +1,9 @@
 /*
 Copyright 2019 The Kubernetes Authors.
-
```
mukhoakash, line 3: Same. Revert unless created by lint fix.

pkg/blob/blob.go
aarongalang: yeah i'm not sure why those ended up there... will revert momentarily

aarongalang: double checked my code, those dashes aren't there, might have been created by lint fixes as you said

andyzhangx (approved):
/lgtm
/approve

PR #744, opened 2022-08-30
feat: always update blobfuse-proxy service in driver init

andyzhangx:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
feat: always update blobfuse-proxy service in driver init

**Which issue(s) this PR fixes**:

Fixes #

**Requirements**:
- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Special notes for your reviewer**:

**Release note**:
```
feat: always update blobfuse-proxy service in driver init
```

PR #761, opened 2022-09-26
fix: restart blobfuse proxy when necessary

cvvz:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Testing result:**

**uninstall old version and install new version:**
image

 
**reinstall the same driver**
image

**Release note**:
```
none
```

### DIFF
--- a/pkg/blobfuse-proxy/init.sh
+++ b/pkg/blobfuse-proxy/init.sh
@@ -40,14 +40,12 @@ if [ -f "/host/usr/bin/blobfuse-proxy" ];then
   if [ "$old" = "$new" ];then
     updateBlobfuseProxy="false"
     echo "no need to update blobfuse-proxy"
-  else
-    rm -rf /host/usr/bin/blobfuse-proxy
-    rm -rf /host/var/lib/kubelet/plugins/blob.csi.azure.com/blobfuse-proxy.sock
   fi
 fi
 
 if [ "$updateBlobfuseProxy" = "true" ];then
   echo "copy blobfuse-proxy...."
+  rm -rf /host/var/lib/kubelet/plugins/blob.csi.azure.com/blobfuse-proxy.sock
   cp /blobfuse-proxy/blobfuse-proxy /host/usr/bin/blobfuse-proxy
   chmod 755 /host/usr/bin/blobfuse-proxy
 fi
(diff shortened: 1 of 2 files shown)


andyzhangx (commented):
/ok-to-test
so this PR already has logic of PR: https://github.com/kubernetes-sigs/blob-csi-driver/pull/744

cvvz:
> so this PR already has logic of PR: https://github.com/kubernetes-sigs/blob-csi-driver/pull/744
Yes, it aims to solve the issue the same as #744 and solve the [comment](https://github.com/kubernetes-sigs/blob-csi-driver/pull/744#discussion_r961385892).

pkg/blobfuse-proxy/init.sh
```diff
@@ -33,25 +33,50 @@ then
...
-if [ ! -f "/host/usr/bin/blobfuse-proxy" ];then
+updateBlobfuseProxy="true"
+if [ -f "/host/usr/bin/blobfuse-proxy" ];then
+  old=$(sha256sum /host/usr/bin/blobfuse-proxy | awk '{print $1}')
+  new=$(sha256sum /blobfuse-proxy/blobfuse-proxy | awk '{print $1}')
+  if [ "$old" = "$new" ];then
+    updateBlobfuseProxy="false"
+    echo "no need to update blobfuse-proxy"
+  else
+    rm -rf /host/usr/bin/blobfuse-proxy
```
andyzhangx, line 44: since it would be overwritten later, no need to do this?

pkg/blobfuse-proxy/init.sh
```diff
@@ -33,25 +33,50 @@ then
...
-if [ ! -f "/host/usr/lib/systemd/system/blobfuse-proxy.service" ];then
+updateService="true"
+if [ -f "/host/usr/lib/systemd/system/blobfuse-proxy.service" ];then
+  old=$(sha256sum /host/usr/lib/systemd/system/blobfuse-proxy.service | awk '{print $1}')
+  new=$(sha256sum /blobfuse-proxy/blobfuse-proxy.service | awk '{print $1}')
+  if [ "$old" = "$new" ];then
+    updateService="false"
+    echo "no need to update blobfuse-proxy.service"
+  else
+    rm -rf /host/usr/lib/systemd/system/blobfuse-proxy.service
```
andyzhangx, line 63: since it would be overwritten later, no need to do this?

pkg/blobfuse-proxy/init.sh
cvvz: fixed

pkg/blobfuse-proxy/init.sh
cvvz: fixed

andyzhangx (approved):
/lgtm
/approve

PR #852, opened 2023-03-20
fix: wait to make sure blobfuse mount complete before bind mount

cvvz:
**What type of PR is this?**

Wait a few seconds to make sure blobfuse mount is successful
Please refer to https://github.com/Azure/azure-storage-fuse/pull/1088 for more details

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/blob/nodeserver.go
```diff
@@ -575,3 +582,20 @@ func (d *Driver) ensureMountPoint(target string, perm os.FileMode) (bool, error)
 	}
 	return !notMnt, nil
 }
+
+func waitForMount(path string, intervel, timeout time.Duration) error {
```
andyzhangx, line 586: pls add a ut for this func, e.g. if `bind mount` to local disk, it should return `notMount as true`

andyzhangx: and also add some logging to check when the blobfuse mount would ready after a few retries.

andyzhangx:
btw, capz failure would be addressed by https://github.com/kubernetes-sigs/blob-csi-driver/pull/853

pkg/blob/nodeserver.go
cvvz: added.

pkg/blob/nodeserver.go
```diff
@@ -575,3 +582,24 @@ func (d *Driver) ensureMountPoint(target string, perm os.FileMode) (bool, error)
...
 }
+
+func (d *Driver) waitForMount(path string, intervel, timeout time.Duration) error {
+	timeAfter := time.After(timeout)
+	timeTick := time.Tick(intervel)
+
+	for {
+		select {
+		case <-timeTick:
+			notMount, err := d.mounter.IsLikelyNotMountPoint(path)
```
andyzhangx, line 593: LGTM in general, is it possible to add a bind mount to local disk path unit test?

cvvz: mount operation needs root permission, which will cause unit test failed locally.

andyzhangx: what about writing a ut case for normal path, it should retrun `true, nil`

cvvz: `IsLikelyNotMountPoint` and `mountinfo.MountedFast` are only support for unix, so the ut in windows and mac os system will fail or cannot compile, so I have to use [fake mounter](https://github.com/kubernetes-sigs/blob-csi-driver/pull/852/files#diff-2ac9427651edf35d972cc03bfea020855df0d4b212b619588756a71d8d875ee9R804).

andyzhangx: I think we could skip windows and mac ut if it does not work, that's ok. It's better to have such test coverage since that's quite fundamental change.

cvvz: fxied, please take another look.

andyzhangx: thanks, pls squash commits

andyzhangx (approved):
/lgtm

andyzhangx:
@cvvz could you cherry pick this PR to release-1.19 and release-1.20? thanks.

PR #857, opened 2023-03-27
feat: support install specific blobfuse version using blobfuse-proxy

andyzhangx:
**What type of PR is this?**
/kind feature
/kind bug

fix: support install specific blobfuse version using blobfuse-proxy

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
fix: support install specific blobfuse version using blobfuse-proxy
```


pkg/blobfuse-proxy/init.sh
```diff
@@ -17,21 +17,54 @@
...
+    # install blobfuse2 with latest version or specific version
+    if [ -z "${BLOBFUSE2_VERSION}" ]; then
+      echo "install blobfuse2 with latest version"
+      pkg_list="${pkg_list} blobfuse2"
+    else
+      pkg_list="${pkg_list} blobfuse2=${BLOBFUSE2_VERSION}"
+    fi
+  fi
+  echo "begin to install ${pkg_list}"
+  $HOST_CMD apt-get install -y $pkg_list
```
cvvz, line 67: We are using fixed version of deb package: https://github.com/kubernetes-sigs/blob-csi-driver/blob/master/pkg/blobplugin/Dockerfile#L36-L37. Is it possible that some of the blobfuse/blobfuse2 version are not included in it?

andyzhangx: yes, that's possible if `packages-microsoft-prod.deb` in the driver is old, and since BLOBFUSE_VERSION can be specified by end user, if it's not supported, then it would return error, so that's ok

PR #861, opened 2023-03-29
test: speed up e2e test by running in parallel

cvvz:
**What type of PR is this?**
/kind test

Currently e2e test is running by `go test` in serial mode, change it to parallel mode to fast the pipeline and detect flaky failure early. 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


cvvz:
Suite will probably be panic if there is spec failed, the problem will be solved either by https://github.com/onsi/ginkgo/pull/1178 or https://github.com/kubernetes/kubernetes/pull/112923 (kubernetes v1.27).

I've handled the panic properly, so it won't cause any real problem, we can still get the logs and other information to dig into the failed specs. However, we should upgrade to v1.27 or upgrade ginkgo after https://github.com/onsi/ginkgo/pull/1178 has been merged.

andyzhangx:
I found there are two driver restart cases, will it affect the test runs?

https://github.com/kubernetes-sigs/blob-csi-driver/blob/07f79d432e1d2ef0fd25f3ba717dc13ca6ea60ef/test/e2e/pre_provisioning_test.go#L337-L345

cvvz:
> I found there are two driver restart cases, will it affect the test runs?

Actually, there is only one place to print out log before restart driver, which is in restart script. The others have been removed.

cvvz:
> what's current parallel num and is there a switch to make it serial run if there is issue?

It's based on the number of cores on the machine. And we can simply remove `-p` opition in ginkgo to switch back to serial.

andyzhangx (approved):
/lgtm

PR #874, opened 2023-04-13
[WIP]test: blobfuse2-2.0.3

andyzhangx:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/blobplugin/Dockerfile
+++ b/pkg/blobplugin/Dockerfile
@@ -34,7 +34,9 @@ ARG ARCH=amd64
 RUN if [ "$ARCH" = "amd64" ] ; then \
   clean-install libcurl4-gnutls-dev && \
   wget -O /blobfuse-proxy/packages-microsoft-prod.deb https://packages.microsoft.com/config/ubuntu/18.04/packages-microsoft-prod.deb && \
-  dpkg -i /blobfuse-proxy/packages-microsoft-prod.deb && apt update && apt install blobfuse blobfuse2 fuse -y && apt remove wget -y; fi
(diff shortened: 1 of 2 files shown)


pkg/blobplugin/Dockerfile
```diff
@@ -34,7 +34,9 @@ ARG ARCH=amd64
 RUN if [ "$ARCH" = "amd64" ] ; then \
   clean-install libcurl4-gnutls-dev && \
   wget -O /blobfuse-proxy/packages-microsoft-prod.deb https://packages.microsoft.com/config/ubuntu/18.04/packages-microsoft-prod.deb && \
-  dpkg -i /blobfuse-proxy/packages-microsoft-prod.deb && apt update && apt install blobfuse blobfuse2 fuse -y && apt remove wget -y; fi
```
vibhansa-msft, line 37: This binary was given for temperory testing so lets not take dependency on this in your regular builds. Also, here binary is pointing to 2.0.3 binary, if we wish to make this generic better we make the binary name generic here, otherwise every release will need a change in this script as well.

andyzhangx: this PR is only for testing the 2.0.3 binary, we won't merge this PR into master branch since it's only for running e2e test.

andyzhangx: and once you have published v2.0.3 into https://packages.microsoft.com/config/ubuntu/18.04/packages-microsoft-prod.deb, there is no change on this project, I only need to change the fixed version(it's now 2.0.2) in AKS vhd image.

PR #892, opened 2023-04-25
fix: install blobfuse package issue in blobfuse-proxy initContainer

cvvz:
**What type of PR is this?**

Fixes #
1. fuse and fuse3 can't be installed at the same time: `fuse3 : Breaks: fuse`
2. blobfuse returned error `error while loading shared libraries: libfuse.so.2: cannot open shared object file: No such file or directory` after using 18.04 packages-microsoft-prod.deb package to install blobfuse on ubuntu 22.04
3. blobfuse2 returned error `error while loading shared libraries: libfuse.so.2: cannot open shared object file: No such file or directory` after using 18.04 packages-microsoft-prod.deb package to install blobfuse2 on ubuntu 22.04
4. mariner os can't use apt-get to install package

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


andyzhangx (approved):
/lgtm

PR #997, opened 2023-08-24
fix: flag panic if a flag is defined after being set

cvvz:
**What type of PR is this?**
/kind bug

golang 1.21 introduced a fix to flag: https://github.com/golang/go/commit/00401835c1d5122a84f0500e48655cd787402f3f, which makes the `Set` fail loudly (panic) if it happened before the definition. Before go1.21 it just fails silently.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
fix: flag panic if a flag is defined after being set
```


andyzhangx:
can you squash commits? thanks.

andyzhangx:
btw, do we need to cherrypick and fix on other csi drivers?

cvvz:
> do we need to cherrypick and fix on other csi drivers?
Other CSI driver doesn't have such usage. So, I think not.

PR #1027, opened 2023-10-07
doc: refine inline volume example

### LINKED ISSUE #1022: Inline volume support for nfs protocol to azure storage account blob

andyzhangx:
**What type of PR is this?**
/kind documentation

doc: refine inline volume example

Fixes #1022

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


calohmn:
> to avoid performance issue, use persistent volume instead of inline volume when numerous pods are accessing the same volume.

@andyzhangx This is a bit confusing to me. Since the inline volume is directly associated with the pod, there is only one pod accessing one inline volume, if I'm not mistaken. Or did you mean "accessing the same storage account/container"?

PR #1094, opened 2023-11-07
Refactor: adopt dependency injection pattern and extract cloud property

MartinForReal:
**What type of PR is this?**
/kind cleanup

adopt dependency injection pattern and extract cloud property from driver factory

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/blob/controllerserver_test.go
+++ b/pkg/blob/controllerserver_test.go
@@ -164,18 +164,6 @@ func TestCreateVolume(t *testing.T) {
 		testFunc func(t *testing.T)
 	}{
 		{
-			name: "invalid create volume req",
-			testFunc: func(t *testing.T) {
-				d := NewFakeDriver()
-				req := &csi.CreateVolumeRequest{}
-				_, err := d.CreateVolume(context.Background(), req)
-				expectedErr := status.Error(codes.InvalidArgument, "CREATE_DELETE_VOLUME")
-				if !reflect.DeepEqual(err, expectedErr) {
-					t.Errorf("actualErr: (%v), expectedErr: (%v)", err, expectedErr)
-				}
-			},
-		},
-		{
 			name: "volume Name missing",
 			testFunc: func(t *testing.T) {
 				d := NewFakeDriver()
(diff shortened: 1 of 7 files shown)


pkg/blobplugin/main.go
```diff
@@ -81,31 +81,34 @@ func handle() {
...
 		EnableAznfsMount:                       *enableAznfsMount,
 		VolStatsCacheExpireInMinutes:           *volStatsCacheExpireInMinutes,
 		SasTokenExpirationMinutes:              *sasTokenExpirationMinutes,
 	}
-	driver := blob.NewDriver(&driverOptions)
+
+	userAgent := blob.GetUserAgent(driverOptions.DriverName, *customUserAgent, *userAgentSuffix)
+	klog.V(2).Infof("driver userAgent: %s", userAgent)
+
+	cloud, err := blob.GetCloudProvider(*kubeconfig, driverOptions.NodeID, *cloudConfigSecretName, *cloudConfigSecretNamespace, userAgent, *allowEmptyCloudConfig, *kubeAPIQPS, *kubeAPIBurst)
```
andyzhangx, line 101: why not put all these into `DriverOptions`? that's the reason why `DriverOptions` exists, it could save parameter num
```
*cloudConfigSecretName, *cloudConfigSecretNamespace, userAgent, *allowEmptyCloudConfig, *kubeAPIQPS, *kubeAPIBurst
```

MartinForReal: It's a little bit complicated to add mock for GetCloudProvider

pkg/blob/controllerserver_test.go
```diff
@@ -893,20 +881,6 @@ func TestDeleteVolume(t *testing.T) {
 				}
 			},
 		},
-		{
```
MartinForReal, line 896: This case is deleted because fake driver has got required capabilities.

pkg/blobplugin/main.go
andyzhangx: but why move part of the options as parameters? this is not consistent

MartinForReal: Should I move other dependencies out of newDriver function? In that case that would be easier to compose driver object and there is no need to implement FakeDriver.

andyzhangx (approved):
/lgtm

pkg/blob/controllerserver_test.go
andyzhangx: what's the result of this test now? `invalid create volume req"` ?

andyzhangx: original test is for testing `invalid create volume req`, not capabilities

pkg/blob/controllerserver_test.go
```diff
@@ -893,20 +881,6 @@ func TestDeleteVolume(t *testing.T) {
...
-			name: "invalid delete volume req",
```
andyzhangx, line 897: could you also add this ut back?

PR #1098, opened 2023-11-08
[cherry-pick: release-1.23] feat: upgrade azcopy to 10.21.2

umagnus:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

[cherry-pick: release-1.23] feat: upgrade azcopy to 10.21.2

**Which issue(s) this PR fixes**:

Fixes #

**Requirements**:
- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Special notes for your reviewer**:

**Release note**:
```
none
```

PR #1102, opened 2023-11-08
[release-1.23] feat: use azcopy for volume cloning

umagnus:
**What type of PR is this?**

/kind feature

[cherry-pick release-1.23] feat: use azcopy for volume cloning

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


.trivyignore
```diff
@@ -0,0 +1,2 @@
+CVE-2023-39325
```
andyzhangx, line 1: pls patch with https://github.com/kubernetes-sigs/blob-csi-driver/pull/1098, so we don't need this file any more.

umagnus: done. combined #1098 in this pr.

PR #1158, opened 2023-12-08
fix: bump azclient and fix signature change

MartinForReal:
**What type of PR is this?**
/kind cleanup

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/blob/controllerserver.go
```diff
@@ -817,7 +817,9 @@ func generateSASToken(accountName, accountKey, storageEndpointSuffix string, exp
 	sasURL, err := serviceClient.GetSASURL(
 		sas.AccountResourceTypes{Object: true, Service: false, Container: true},
 		sas.AccountPermissions{Read: true, List: true, Write: true},
-		sas.AccountServices{Blob: true}, time.Now(), time.Now().Add(time.Duration(expiryTime)*time.Minute))
```
andyzhangx, line 820: you are missing `sas.AccountServices{Blob: true}, time.Now()` here.
cc @umagnus

MartinForReal: the client is bumped to 1.5

andyzhangx: so `sas.AccountServices` is removed in 1.5 ? @umagnus can you check whether there would be issue?

MartinForReal:
<html>
<body>

/test pull-blob-csi-driver-e2e
--


<br class="Apple-interchange-newline">
</body>
</html>

pkg/blob/controllerserver.go
umagnus: in v1.1 is always blob, I think it wouldn't be issue
https://github.com/Azure/azure-sdk-for-go/blob/48a06486ac095e7bd94e3527ca5c43465b87e102/sdk/storage/azblob/sas/account.go#L64C17-L64C17

andyzhangx (approved):
/lgtm

PR #1169, opened 2023-12-13
Refactor: migrate blobclient to track2 sdk and fix ut cases

MartinForReal:
**What type of PR is this?**
/kind cleanup

Refactor: migrate blobclient to track2 sdk and fix ut cases

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

gomock controller is not closed after ut test is complete. This may miss unexpected function call error.

**Release note**:
```
none
```


pkg/blobplugin/main.go
```diff
@@ -83,8 +84,41 @@ func handle() {
...
 	klog.V(2).Infof("cloud: %s, location: %s, rg: %s, VnetName: %s, VnetResourceGroup: %s, SubnetName: %s", cloud.Cloud, cloud.Location, cloud.ResourceGroup, cloud.VnetName, cloud.VnetResourceGroup, cloud.SubnetName)
-
-	driver := blob.NewDriver(&driverOptions, kubeClient, cloud)
+	authProvider, err := azclient.NewAuthProvider(&cloud.ARMClientConfig, &cloud.AzureAuthConfig.AzureAuthConfig)
+	if err != nil {
+		klog.Fatalf("failed to create auth provider: %v", err)
+	}
+	multiTenantAuth, err := authProvider.GetMultiTenantIdentity()
+	if err != nil {
+		klog.Warningf("failed to create multi tenant auth: %v", err)
```
andyzhangx, line 93: should be Fatal error here?

pkg/blobplugin/main.go
```diff
@@ -83,8 +84,41 @@ func handle() {
...
+	}
+	clientFactory, err := azclient.NewClientFactory(&azclient.ClientFactoryConfig{
+		SubscriptionID: cloud.SubscriptionID,
+	}, &cloud.ARMClientConfig, cred)
+	if err != nil {
+		klog.Fatalf("failed to create client factory: %v", err)
+	}
+	networkAuth, err := authProvider.GetNetworkAzIdentity()
+	if err != nil {
+		klog.Warningf("failed to create network auth: %v", err)
```
andyzhangx, line 110: should be Fatal error here?

pkg/blobplugin/main.go
```diff
@@ -83,8 +84,41 @@ func handle() {
 		klog.Fatalf("failed to get Azure Cloud Provider, error: %v", err)
 	}
 	klog.V(2).Infof("cloud: %s, location: %s, rg: %s, VnetName: %s, VnetResourceGroup: %s, SubnetName: %s", cloud.Cloud, cloud.Location, cloud.ResourceGroup, cloud.VnetName, cloud.VnetResourceGroup, cloud.SubnetName)
-
-	driver := blob.NewDriver(&driverOptions, kubeClient, cloud)
+	authProvider, err := azclient.NewAuthProvider(&cloud.ARMClientConfig, &cloud.AzureAuthConfig.AzureAuthConfig)
+	if err != nil {
+		klog.Fatalf("failed to create auth provider: %v", err)
```
andyzhangx, line 89: originally the driver supports empty auth, what if user does not provide cloud provider config?

andyzhangx (commented):
@cvvz could you also take a look at this PR? thanks.

andyzhangx:
all the 3 Azure CSI drivers supports empty cloud provider config, e.g.
https://github.com/kubernetes-sigs/blob-csi-driver/blob/703d8dd45b0b06178a8cb50b1822fa77005067d1/pkg/blob/azure.go#L97-L98

we need to make sure if there is no cloud provider config provided, this driver won't crash.

pkg/blobplugin/main.go
MartinForReal: the code has changed. here we will check if multi-tenant is enabled or not.

pkg/blobplugin/main.go
MartinForReal: it will never reach here. I'll update authProvider.

pkg/blobplugin/main.go
MartinForReal: the code has changed. the clientFactory is nil if cred is nil.

MartinForReal:
If the cloud config is nil, the blob client is nil. How do we check volume capabilities in this scenario? @andyzhangx

cvvz (commented):
> If the cloud config is nil, the blob client is nil. How do we check volume capabilities in this scenario? @andyzhangx  

I believe k8s and sidecars actually never call `ValidateVolumeCapabilities`, although CSI spec requires CSI driver to implement it...

andyzhangx:
> > If the cloud config is nil, the blob client is nil. How do we check volume capabilities in this scenario? @andyzhangx
> 
> I believe k8s and sidecars actually never call `ValidateVolumeCapabilities`, although CSI spec requires CSI driver to implement it...

If the cloud config is nil, then user should provide secrets(bring your own account key scenario), `ValidateVolumeCapabilities` would work well with secrets.

cvvz:
> If the cloud config is nil, then user should provide secrets(bring your own account key scenario), `ValidateVolumeCapabilities` would work well with secrets.

Yes, but it can be very easy to make csi driver crash (rather than throw error) once csi driver walk into the code that use `d.cloud.BlobClient`.

andyzhangx:
> > If the cloud config is nil, then user should provide secrets(bring your own account key scenario), `ValidateVolumeCapabilities` would work well with secrets.
> 
> Yes, but it can be very easy to make csi driver crash (rather than throw error) once csi driver walk into the code that use `d.cloud.BlobClient`.

then we should check whether `d.cloud.BlobClient` is nil, if it's nil, then return error directly instead of crashing.

andyzhangx:
> > If the cloud config is nil, then user should provide secrets(bring your own account key scenario), `ValidateVolumeCapabilities` would work well with secrets.
> 
> Yes, but it can be very easy to make csi driver crash (rather than throw error) once csi driver walk into the code that use `d.cloud.BlobClient`.

@cvvz could you take a look at this PR? any concern about blobclient issue?

andyzhangx (approved):
/lgtm

PR #1173, opened 2023-12-14
fix: support install blobfuse proxy on other linux distributions

### LINKED ISSUE #1172: Something is off with blobfuse-proxy on 1.23.2 release
**What happened**: csi blob mounts won't mount anymore on release 1.23.2 with blobfuse-proxy enabled (new install or upgrades from 1.23.1 with blobfuse-proxy previously disabled)

**What you expected to happen**: csi blob mount is mounted succesfully !

**How to reproduce it**: inline method, try to mount a blob on a pod

**Anything else we need to know?**: 1.23.1 was working fine..

Error message:
```
MountVolume.SetUp failed for volume "azure-blob" : rpc error: code = Internal desc = Mount failed with error: context deadline exceeded, output:
```

Here are my findings:

```
1. Install csi driver 1.23.1 without blobfuse proxy (mount works fine)
  1. Upgrade to 1.23.2 With blobfuse-proxy (mount does not work)
1. Install csi driver 1.23.1 with blobfuse proxy (mount works fine)
  1. Upgrade to 1.23.2 with blobfuse proxy (Mount works fine)
1. Install csi driver 1.23.2 with blobfuse proxy (Mount does not work)
1. Install csi driver 1.23.2 without blobfuse proxy (Mount works fine)
```

I am very confused on why this is happening ! All tests are with new pods after the installs or upgrade

We're on Rocky Linux 8.8

cvvz:
**What type of PR is this?**
/kind bug

fix: support install blobfuse proxy on other linux distributions

Fixes #1172

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
fix: support install blobfuse proxy on other linux distributions
```


cvvz:
The installation scripts of mariner and ubuntu and any other Linux distributions basically have the same logical, so I use `install-proxy.sh` as the entry point for all the Linux distributions except for coreos.

andyzhangx:
@cvvz thanks for the quick fix! we did not realize that this csi driver is also used on other distros, in the future, I think we need to add some e2e tests or basic functional tests on other distros.

andyzhangx (approved):
/lgtm

PR #1309, opened 2024-03-26
fix: strip service account token

cvvz:
**What type of PR is this?**
/kind cleanup

strip service account token to avoid security risk.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
fix: strip service account token
```

### DIFF
--- a/pkg/csi-common/utils.go
+++ b/pkg/csi-common/utils.go
@@ -102,7 +102,7 @@ func getLogLevel(method string) int32 {
 func LogGRPC(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
 	level := klog.Level(getLogLevel(info.FullMethod))
 	klog.V(level).Infof("GRPC call: %s", info.FullMethod)
-	klog.V(level).Infof("GRPC request: %s", stripServiceAccountToken(protosanitizer.StripSecrets(req)))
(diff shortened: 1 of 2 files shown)


pkg/csi-common/utils.go
```diff
@@ -111,3 +112,30 @@ func LogGRPC(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, h
...
+	if err != nil || parsed == nil {
+		return req.String()
+	}
+
+	volumeContext, ok := parsed["volume_context"].(map[string]interface{})
+	if !ok {
+		return req.String()
+	}
+
+	if _, ok := volumeContext["csi.storage.k8s.io/serviceAccount.tokens"]; !ok {
```
andyzhangx, line 129: let's make this func more generic, set `csi.storage.k8s.io/serviceAccount.tokens` as a parameter, thanks.

andyzhangx: rename as `stripSecrect(secretFieldNames []string)`

cvvz: ok, please take another look

cvvz: > rename as stripSecrect(secretFieldNames []string)

This will be conflict with the `StripSecrets` function in csi lib

andyzhangx (approved):
/lgtm
/retest

PR #1425, opened 2024-06-14
fix: force copy blobfuse2 binary to avoid Text file busy error

### LINKED ISSUE #1405: "install-blobfuse-proxy" container fails on OpenShift (error copying blobfuse2 - "Text file busy")
**What happened**:
We are using blob-csi-driver-v1.23.2 on OpenShift.
We are frequently seeing `csi-blob-node` pods to not get ready because of an error in the `install-blobfuse-proxy` container.
The error is about copying the blobfuse2 executable:
`cp: cannot create regular file '/host/usr/local/bin/blobfuse2': Text file busy`

Full log:
```
+ INSTALL_BLOBFUSE_PROXY=true
+ DISABLE_UPDATEDB=true
+ SET_MAX_OPEN_FILE_NUM=true
+ SET_READ_AHEAD_SIZE=true
+ READ_AHEAD_KB=15380
+ HOST_CMD=nsenter --mount=/proc/1/ns/mnt
+ nsenter --mount=/proc/1/ns/mnt cat /etc/os-release
+ grep ^ID=
+ cut -d= -f2
+ tr -d "
+ DISTRIBUTION=rhcos
+ nsenter --mount=/proc/1/ns/mnt uname -m
+ ARCH=x86_64
+ echo Linux distribution: rhcos, Arch: x86_64
Linux distribution: rhcos, Arch: x86_64
+ . ./blobfuse-proxy/install-proxy-rhcos.sh
+ set -xe
+ [ true = true ]
+ echo copy blobfuse2....
+ copy blobfuse2....
cp /usr/bin/blobfuse2 /host/usr/local/bin/blobfuse2
cp: cannot create regular file '/host/usr/local/bin/blobfuse2': Text file busy
```

It looks like a fix similar to #776 needs to be applied to `pkg/blobfuse-proxy/install-proxy-rhcos.sh`, deleting `/host/usr/local/bin/blobfuse2` first before copying the blobfuse2 executable:
https://github.com/kubernetes-sigs/blob-csi-driver/blob/7e434dbc816a4a390dd6f58390e7c82cfa8d4e63/pkg/blobfuse-proxy/install-proxy-rhcos.sh#L19-L24

`install-blobfuse-proxy` container startup without error.

**How to reproduce it**:
Using the blob-csi-driver on OpenShift and restarting a `csi-blob-node` pod.

**Environment**:

- CSI Driver version: `1.23.2`
- Kubernetes version (use `kubectl version`): `v1.25.11+1485cc9`
- OS (e.g. from /etc/os-release): `rhcos`
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

cvvz:
**What type of PR is this?**
/kind bug

fix: force copy blobfuse2 binary to avoid Text file busy error

Fixes #1405

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
fix: force copy blobfuse2 binary to avoid Text file busy error
```

### DIFF
--- a/pkg/blobfuse-proxy/install-proxy-rhcos.sh
+++ b/pkg/blobfuse-proxy/install-proxy-rhcos.sh
@@ -17,14 +17,25 @@
 set -xe
 
 # in coreos, we could just copy the blobfuse2 binary to /usr/local/bin/blobfuse2
+updateBlobfuse2="true"
 if [ "${INSTALL_BLOBFUSE}" = "true" ] || [ "${INSTALL_BLOBFUSE2}" = "true" ]
 then
-  old=$(sha256sum /host/usr/local/bin/blobfuse2 | awk '{print $1}')
-  new=$(sha256sum /usr/bin/blobfuse2 | awk '{print $1}')
-  if [ "$old" != "$new" ];then
-    echo "copy blobfuse2...."
-    cp /usr/bin/blobfuse2 /host/usr/local/bin/blobfuse2 --force
+  if [ -f "/host/usr/local/bin/blobfuse2" ];then
+    old=$(sha256sum /host/usr/local/bin/blobfuse2 | awk '{print $1}')
+    new=$(sha256sum /usr/bin/blobfuse2 | awk '{print $1}')
+    if [ "$old" = "$new" ];then
+      updateBlobfuse2="false"
+      echo "no need to update blobfuse2 since no change"
+    fi
   fi
+else
+  echo "no need to install blobfuse2"
+  updateBlobfuse2="false"
+fi
+if [ "$updateBlobfuse2" = "true" ];then
+  echo "copy blobfuse2...."
+  cp /usr/bin/blobfuse2 /host/usr/local/bin/blobfuse2 --force
+  chmod 755 /host/usr/local/bin/blobfuse2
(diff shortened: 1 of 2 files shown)


pkg/blobfuse-proxy/install-proxy-rhcos.sh
```diff
@@ -17,10 +17,25 @@
 set -xe
 
 # in coreos, we could just copy the blobfuse2 binary to /usr/local/bin/blobfuse2
+updateBlobfuse2="true"
```
andyzhangx, line 20: updateBlobfuse2 should be true only when INSTALL_BLOBFUSE2 is true

cvvz: It will be set to false when INSTALL_BLOBFUSE2 is not true in line 33.

cvvz: updateBlobfuse2 will be set as true only when INSTALL_BLOBFUSE2 is true and blobfuse2 does not exist or the old binary file is not the same as the new file.

andyzhangx (approved):
/lgtm

PR #1470, opened 2024-07-11
feat: add allowSharedKeyAccess parameter

### LINKED ISSUE #1462: Dynamically provisioned storage account with shared key access disabled
I'd like to have dynamically provisioned storage account have shared key access disabled.

 Introduce new driver parameter that will control storage account shared key access setting (AccountOptions.AllowSharedKeyAccess)

NA

**Additional context**
https://learn.microsoft.com/en-us/azure/storage/common/shared-key-authorization-prevent

specialforest:
**What type of PR is this?**
/kind feature

Allows to provision dynamic storage accounts with shared key access disabled.

Fixes #1462.

**Requirements**:
  
- [x] includes documentation
- [x] adds unit tests

Tested on AKS cluster by mounting dynamic PVC to a pod and writing a file.
Storage class spec:
```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: azureblob-fuse
reclaimPolicy: Delete
volumeBindingMode: Immediate
provisioner: blob.csi.azure.com
allowVolumeExpansion: true
parameters:
  skuName: Premium_LRS
  allowSharedKeyAccess: "false"
  storeAccountKey: "false"
  AzureStorageAuthType: msi
  AzureStorageIdentityClientID: <redacted>
```

**Release note**:
```
none
```

### DIFF
--- a/pkg/blob/controllerserver.go
+++ b/pkg/blob/controllerserver.go
@@ -105,6 +105,8 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	var err error
 	// set allowBlobPublicAccess as false by default
 	allowBlobPublicAccess := pointer.Bool(false)
+	// set allowBlobPublicAccess as true by default
+	allowSharedKeyAccess := pointer.Bool(true)
 
 	containerNameReplaceMap := map[string]string{}
 
@@ -170,6 +172,10 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		case allowBlobPublicAccessField:
 			if strings.EqualFold(v, trueValue) {
 				allowBlobPublicAccess = pointer.Bool(true)
+			}
+		case allowSharedKeyAccessField:
+			if strings.EqualFold(v, falseValue) {
+				allowSharedKeyAccess = pointer.Bool(false)
 			}
 		case requireInfraEncryptionField:
 			if strings.EqualFold(v, trueValue) {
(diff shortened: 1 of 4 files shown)


specialforest:
Need to add a unit test.

andyzhangx:
`allowSharedKeyAccess` parameter needs to be added in azure cloud provider first

andyzhangx:
> `allowSharedKeyAccess` parameter needs to be added in azure cloud provider first

nvm, `allowSharedKeyAccess` is already added in azure cloud provider

pkg/blob/controllerserver.go
```diff
@@ -310,6 +316,10 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		storageEndpointSuffix = d.getStorageEndPointSuffix()
 	}
 
+	if storeAccountKey && !pointer.BoolDeref(allowSharedKeyAccess, false) {
```
umagnus, line 319: I think here should be `!pointer.BoolDeref(allowSharedKeyAccess, true)` since if we didn't give allowSharedKeyAccess the default is true

andyzhangx: yes, the default value is true

specialforest: Technically allowSharedKeyAccess is never nil here but let me refactor the code a little to make it more straightforward.

andyzhangx (commented):
/ok-to-test

andyzhangx (approved):
/lgtm

PR #1490, opened 2024-07-16
feat: add metadata tag on the storage account created by the driver

umagnus:
**What type of PR is this?**

/kind feature

feat: add metadata in container creating by csi driver, `createdBy: blob.csi.azure.com`.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

test result:
The first used dataplane api
image
image

**Release note**:
```
none
```

### DIFF
--- a/pkg/blob/controllerserver.go
+++ b/pkg/blob/controllerserver.go
@@ -61,6 +61,8 @@ const (
 	MSI                             = "MSI"
 	SPN                             = "SPN"
 	authorizationPermissionMismatch = "AuthorizationPermissionMismatch"
+
+	createdByMetadata = "createdBy"
 )
 
 // CreateVolume provisions a volume
@@ -691,11 +693,13 @@ func (d *Driver) CreateBlobContainer(ctx context.Context, subsID, resourceGroupN
 			if getErr != nil {
 				return true, getErr
 			}
+			container.Metadata = map[string]string{createdByMetadata: d.Name}
 			_, err = container.CreateIfNotExists(&azstorage.CreateContainerOptions{Access: azstorage.ContainerAccessTypePrivate})
 		} else {
 			blobContainer := armstorage.BlobContainer{
 				ContainerProperties: &armstorage.ContainerProperties{
 					PublicAccess: to.Ptr(armstorage.PublicAccessNone),
+					Metadata:     map[string]*string{createdByMetadata: to.Ptr(d.Name)},
 				},
 			}
 			var blobClient blobcontainerclient.Interface


pkg/blob/controllerserver.go
```diff
@@ -691,11 +698,17 @@ func (d *Driver) CreateBlobContainer(ctx context.Context, subsID, resourceGroupN
...
 			_, err = container.CreateIfNotExists(&azstorage.CreateContainerOptions{Access: azstorage.ContainerAccessTypePrivate})
 		} else {
+			metadataPtr := make(map[string]*string)
+			for k, v := range metadata {
+				metadataPtr[k] = to.Ptr(v)
+			}
 			blobContainer := armstorage.BlobContainer{
 				ContainerProperties: &armstorage.ContainerProperties{
 					PublicAccess: to.Ptr(armstorage.PublicAccessNone),
+					Metadata:     metadataPtr,
```
andyzhangx, line 711: let's add a const tag: `createdBy`: driverOptions.DriverName in metadata, don't need to expose to the external user

umagnus: added it

umagnus: fix it to not expose to the cx

pkg/blob/controllerserver.go
```diff
@@ -691,11 +693,17 @@ func (d *Driver) CreateBlobContainer(ctx context.Context, subsID, resourceGroupN
 			if getErr != nil {
 				return true, getErr
 			}
+			metadata := make(map[string]string)
+			metadata[createdByMetadata] = d.Name
+			container.Metadata = metadata
```
andyzhangx, line 698: container.Metadata = map[string]string{createdByMetadata: d.Name}

pkg/blob/controllerserver.go
```diff
@@ -691,11 +693,17 @@ func (d *Driver) CreateBlobContainer(ctx context.Context, subsID, resourceGroupN
...
 			_, err = container.CreateIfNotExists(&azstorage.CreateContainerOptions{Access: azstorage.ContainerAccessTypePrivate})
 		} else {
+			metadataPtr := make(map[string]*string)
+			metadataPtr[createdByMetadata] = to.Ptr(d.Name)
 			blobContainer := armstorage.BlobContainer{
 				ContainerProperties: &armstorage.ContainerProperties{
 					PublicAccess: to.Ptr(armstorage.PublicAccessNone),
+					Metadata:     metadataPtr,
```
andyzhangx, line 706: Metadata: map[string]*string{createdByMetadata: to.Ptr(d.Name)}

pkg/blob/controllerserver.go
umagnus: done

pkg/blob/controllerserver.go
umagnus: done

andyzhangx (approved):
/lgtm

PR #1504, opened 2024-07-19
fix: clone volume content to requested volume

### LINKED ISSUE #1495: Volume Cloning creates Blob Container in wrong Storage Account
I am trying to use the CSI Volume Cloning feature to copy a blob volume from a source storage account with `Blob Data Reader` into a destination storage account with `Blob Data Contributor`.

This is desirable in order to dynamically provision a copy of a blob container from a read-only global source storage account into a read-write cluster adjacent regional destination storage account, which can then be automatically reclaimed when no longer needed.

It appears that when cloning a blob volume, the controller server mistakenly attempts to provision the new blob container in the source storage account instead of the destination storage account.

The system identity does not have the permissions to write to the source storage and fails.

I am including a lightly anonymized copy of the logs and kubernetes resources for repro.

**What happened**:
```
controller.go:1366] provision "datacopy" class "datacopy": started

event.go:298] Event(External provisioner is provisioning volume for claim "datacopy"

utils.go:104] GRPC call: /csi.v1.Controller/CreateVolume

utils.go:105] GRPC request: {"capacity_range":{"required_bytes":274877906944},"name":"pvc-f0b58822-7342-4933-85b1-17ecf2ea94a6","parameters":{"csi.storage.k8s.io/pv/name":"pvc-f0b58822-7342-4933-85b1-17ecf2ea94a6","csi.storage.k8s.io/pvc/name":"data","protocol":"fuse2","resourceGroup":"resourceGroupDatacopy","storageAccount":"storageAccountDatacopy","storeAccountKey":"false"},"volume_capabilities":[{"AccessType":{"Mount":{"mount_flags":["--cancel-list-on-mount-seconds=0","-o allow_other","-o attr_timeout=120","-o entry_timeout=120","-o negative_timeout=120"]}},"access_mode":{"mode":3}}],"volume_content_source":{"Type":{"Volume":{"volume_id":"resourceGroupDatasource#storageAccountDatasource#containerNameDatasource"}}}}

controllerserver.go:802] use system-assigned managed identity to authorize azcopy

controllerserver.go:750] begin to copy blob container datasource to pvc-f0b58822-7342-4933-85b1-17ecf2ea94a6

controllerserver.go:760] copy blob container datasource to pvc-f0b58822-7342-4933-85b1-17ecf2ea94a6

controllerserver.go:767] CopyBlobContainer(resourceGroupDatasource, storageAccountDatasource, https://storageAccountDatasource.blob.core.windows.net/pvc-f0b58822-7342-4933-85b1-17ecf2ea94a6) failed with error(exit status 1): 

INFO: Scanning...
INFO: Login with identity succeeded.
INFO: Authenticating to destination using Azure AD
INFO: Authenticating to source using Azure AD
INFO: Failed to create one or more destination container(s). Your transfers may still succeed if the container already exists.
INFO: Authentication failed, it is either not correct, or expired, or does not have the correct permission PUT https://storageAccountDatasource.blob.core.windows.net/pvc-f0b58822-7342-4933-85b1-17ecf2ea94a6/file4.bin
```

A dynamic blob container should have been provisioned in the storage account specified in the datacopy storage class.

**How to reproduce it**:

See example kubernetes specs below.

Functioning volume cloning is a highly desirable feature.

**Environment**:
- CSI Driver version: 
mcr.microsoft.com/oss/kubernetes-csi/blob-csi:v1.23.4
- Kubernetes version (use `kubectl version`): 
Client Version: v1.29.2
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.28.10

```kubernetes
# datasource volume for existing blob container in existing storage account on which the aks system identity has Blob Data Reader
apiVersion: v1
kind: PersistentVolume
metadata:
  name: datasource
spec:
  capacity:
    storage: 1
  accessModes:
  - ReadOnlyMany
  persistentVolumeReclaimPolicy: Retain
  storageClassName: ''
  claimRef:
    name: datasource
    namespace: blob-csi-test
  csi:
    driver: blob.csi.azure.com
    volumeHandle: resourceGroupDatasource#storageAccountDatasource#containerNameDatasource
    volumeAttributes:
      protocol: fuse2
      storeAccountKey: 'false'
      resourceGroup: resourceGroupDatasource
      storageAccount: storageAccountDatasource
      containerName: containerNameDatasource

---
# datasource volume claim
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: datasource
spec:
  accessModes:
    - ReadOnlyMany
  resources:
    requests:
      storage: 1
  storageClassName: ''
  volumeName: datasource

---
# storage class for dynamic blob container in existing storage account on which the aks system identity has Blob Data Contributor
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: datacopy
provisioner: blob.csi.azure.com
parameters:
  protocol: fuse2
  storeAccountKey: 'false'
  resourceGroup: resourceGroupDatacopy
  storageAccount: storageAccountDatacopy
reclaimPolicy: Delete
volumeBindingMode: Immediate
allowVolumeExpansion: true
mountOptions:
  - --cancel-list-on-mount-seconds=0
  - -o allow_other
  - -o attr_timeout=120
  - -o entry_timeout=120
  - -o negative_timeout=120

---
# datacopy volume claim
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: datacopy
spec:
  accessModes:
  - ReadOnlyMany
  storageClassName: datacopy
  resources:
    requests:
      storage: 256Gi
  dataSource:
    kind: PersistentVolumeClaim
    name: datasource

---
# datareader to consume the datacopy
apiVersion: apps/v1
kind: Deployment
metadata:
  name: datareader
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/name: datareader
  template:
    metadata:
      labels:
        app.kubernetes.io/name: datareader
    spec:
      nodeSelector:
        kubernetes.io/os: linux
      volumes:
      - name: data
        persistentVolumeClaim:
          claimName: datacopy
      containers:
      - name: datareader
        image: mcr.microsoft.com/cbl-mariner/base/core:2.0
        volumeMounts:
        - name: data
          mountPath: /vol/data
        command:
          - /bin/bash
          - -c
          - |-
            ls -alR /vol
            sleep 3600s
            exit 0
```

acortelyou:
/kind bug

`CreateVolume()` now follows the established volume provisioning logic when cloning volume content.

This ensures the created volume has the expected configuration, metadata and content.

- Resolves issue #1495
- Prevents SAS tokens from being logged in clear text during volume cloning
- Avoids duplication of azcopy jobs when `containerName` or `containerNamePrefix` are specified

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

- Volumes are now provisioned correctly when cloning

### DIFF
--- a/pkg/blob/controllerserver.go
+++ b/pkg/blob/controllerserver.go
@@ -81,18 +81,11 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
-	if acquired := d.volumeLocks.TryAcquire(volName); !acquired {
-		// logging the job status if it's volume cloning
-		if req.GetVolumeContentSource() != nil {
-			jobState, percent, err := d.azcopy.GetAzcopyJob(volName, []string{})
-			return nil, status.Errorf(codes.Aborted, volumeOperationAlreadyExistsWithAzcopyFmt, volName, jobState, percent, err)
-		}
-		return nil, status.Errorf(codes.Aborted, volumeOperationAlreadyExistsFmt, volName)
-	}
-	defer d.volumeLocks.Release(volName)
-
 	volSizeBytes := int64(req.GetCapacityRange().GetRequiredBytes())
 	requestGiB := int(util.RoundUpGiB(volSizeBytes))
+
+	volContentSource := req.GetVolumeContentSource()
+	secrets := req.GetSecrets()
 
 	parameters := req.GetParameters()
 	if parameters == nil {
@@ -351,10 +344,31 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		GetLatestAccountKey:             getLatestAccountKey,
 	}
 
+	containerName = replaceWithMap(containerName, containerNameReplaceMap)
+	validContainerName := containerName
+	if validContainerName == "" {
+		validContainerName = volName
+		if containerNamePrefix != "" {
+			validContainerName = containerNamePrefix + "-" + volName
+		}
+		validContainerName = getValidContainerName(validContainerName, protocol)
+		setKeyValueInMap(parameters, containerNameField, validContainerName)
+	}
+
+	if acquired := d.volumeLocks.TryAcquire(volName); !acquired {
+		// logging the job status if it's volume cloning
+		if volContentSource != nil {
+			jobState, percent, err := d.azcopy.GetAzcopyJob(validContainerName, []string{})
+			return nil, status.Errorf(codes.Aborted, volumeOperationAlreadyExistsWithAzcopyFmt, volName, jobState, percent, err)
+		}
+		return nil, status.Errorf(codes.Aborted, volumeOperationAlreadyExistsFmt, volName)
+	}
+	defer d.volumeLocks.Release(volName)
+
 	var volumeID string
 	requestName := "controller_create_volume"
-	if req.GetVolumeContentSource() != nil {
-		switch req.VolumeContentSource.Type.(type) {
+	if volContentSource != nil {
+		switch volContentSource.Type.(type) {
 		case *csi.VolumeContentSource_Snapshot:
 			requestName = "controller_create_volume_from_snapshot"
 		case *csi.VolumeContentSource_Volume:
@@ -367,9 +381,39 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		mc.ObserveOperationWithResult(isOperationSucceeded, VolumeID, volumeID)
 	}()
 
+	var srcAzcopyAuthEnv []string
+	var srcSubscriptionID, srcResourceGroupName, srcAccountName, srcContainerName, srcPath, srcAccountSASToken string
+	if volContentSource != nil {
+		switch volContentSource.Type.(type) {
+		case *csi.VolumeContentSource_Snapshot:
+			return nil, status.Errorf(codes.InvalidArgument, "VolumeContentSource Snapshot is not yet implemented")
+		case *csi.VolumeContentSource_Volume:
+			var srcVolumeID string
+			if volContentSource.GetVolume() != nil {
+				srcVolumeID = volContentSource.GetVolume().GetVolumeId()
+			}
+			srcResourceGroupName, srcAccountName, srcContainerName, _, srcSubscriptionID, err = GetContainerInfo(srcVolumeID)
+			if err != nil {
+				return nil, status.Error(codes.NotFound, err.Error())
+			}
+			srcAccountOptions := &azure.AccountOptions{
+				Name:                srcAccountName,
+				SubscriptionID:      srcSubscriptionID,
+				ResourceGroup:       srcResourceGroupName,
+				GetLatestAccountKey: getLatestAccountKey,
+			}
+			srcAccountSASToken, srcAzcopyAuthEnv, err = d.getAzcopyAuth(ctx, srcAccountName, "", storageEndpointSuffix, srcAccountOptions, secrets, secretName, secretNamespace)
+			if err != nil {
+				return nil, status.Errorf(codes.Internal, "failed to getAzcopyAuth on account(%s) rg(%s), error: %v", accountOptions.Name, accountOptions.ResourceGroup, err)
+			}
+			srcPath = fmt.Sprintf("https://%s.blob.%s/%s", srcAccountName, storageEndpointSuffix, srcContainerName)
+		default:
+			return nil, status.Errorf(codes.InvalidArgument, "VolumeContentSource is not recognized: %v", volContentSource)
+		}
+	}
+
 	var accountKey string
 	accountName := account
-	secrets := req.GetSecrets()
 	if len(secrets) == 0 && accountName == "" {
 		if v, ok := d.volMap.Load(volName); ok {
 			accountName = v.(string)
@@ -423,30 +467,25 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		secrets = createStorageAccountSecret(accountName, accountKey)
 	}
 
-	// replace pv/pvc name namespace metadata in subDir
-	containerName = replaceWithMap(containerName, containerNameReplaceMap)
-	validContainerName := containerName
-	if validContainerName == "" {
-		validContainerName = volName
-		if containerNamePrefix != "" {
-			validContainerName = containerNamePrefix + "-" + volName
+	dstAzcopyAuthEnv := srcAzcopyAuthEnv
+	dstAccountSASToken := srcAccountSASToken
+	if volContentSource != nil {
+		if srcSubscriptionID != subsID || srcResourceGroupName != resourceGroup || srcAccountName != accountName {
+			if dstAccountSASToken, dstAzcopyAuthEnv, err = d.getAzcopyAuth(ctx, accountName, accountKey, storageEndpointSuffix, accountOptions, secrets, secretName, secretNamespace); err != nil {
+				return nil, status.Errorf(codes.Internal, "failed to getAzcopyAuth on account(%s) rg(%s), error: %v", accountOptions.Name, accountOptions.ResourceGroup, err)
+			}
 		}
-		validContainerName = getValidContainerName(validContainerName, protocol)
-		setKeyValueInMap(parameters, containerNameField, validContainerName)
 	}
 
-	if req.GetVolumeContentSource() != nil {
-		accountSASToken, authAzcopyEnv, err := d.getAzcopyAuth(ctx, accountName, accountKey, storageEndpointSuffix, accountOptions, secrets, secretName, secretNamespace)
-		if err != nil {
-			return nil, status.Errorf(codes.Internal, "failed to getAzcopyAuth on account(%s) rg(%s), error: %v", accountOptions.Name, accountOptions.ResourceGroup, err)
-		}
-		if err := d.copyVolume(req, accountSASToken, authAzcopyEnv, validContainerName, storageEndpointSuffix); err != nil {
+	klog.V(2).Infof("begin to create container(%s) on account(%s) type(%s) subsID(%s) rg(%s) location(%s) size(%d)", validContainerName, accountName, storageAccountType, subsID, resourceGroup, location, requestGiB)
+	if err := d.CreateBlobContainer(ctx, subsID, resourceGroup, accountName, validContainerName, secrets); err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to create container(%s) on account(%s) type(%s) rg(%s) location(%s) size(%d), error: %v", validContainerName, accountName, storageAccountType, resourceGroup, location, requestGiB, err)
+	}
+
+	if volContentSource != nil {
+		dstPath := fmt.Sprintf("https://%s.blob.%s/%s", accountName, storageEndpointSuffix, validContainerName)
+		if err := d.copyBlobContainer(dstAzcopyAuthEnv, srcPath, srcAccountSASToken, dstPath, dstAccountSASToken, validContainerName); err != nil {
 			return nil, err
-		}
-	} else {
-		klog.V(2).Infof("begin to create container(%s) on account(%s) type(%s) subsID(%s) rg(%s) location(%s) size(%d)", validContainerName, accountName, storageAccountType, subsID, resourceGroup, location, requestGiB)
-		if err := d.CreateBlobContainer(ctx, subsID, resourceGroup, accountName, validContainerName, secrets); err != nil {
-			return nil, status.Errorf(codes.Internal, "failed to create container(%s) on account(%s) type(%s) rg(%s) location(%s) size(%d), error: %v", validContainerName, accountName, storageAccountType, resourceGroup, location, requestGiB, err)
 		}
 	}
 
@@ -488,7 +527,7 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 			VolumeId:      volumeID,
 			CapacityBytes: req.GetCapacityRange().GetRequiredBytes(),
 			VolumeContext: parameters,
-			ContentSource: req.GetVolumeContentSource(),
+			ContentSource: volContentSource,
 		},
 	}, nil
 }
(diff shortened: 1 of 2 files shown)


andyzhangx:
@acortelyou I have pushed a new commit(https://github.com/kubernetes-sigs/blob-csi-driver/pull/1504/commits/b40c5f9a521cac68574d3d3c3307d5907a53e7d5), only copy volume to different account if storageAccount is specified

acortelyou:
Is some code missing from the commit? Or did you want me to implement?

andyzhangx:
> Is some code missing from the commit? Or did you want me to implement?

@acortelyou I have made the changes already, thx

acortelyou:
I have some concerns about these changes, will review more thoroughly tomorrow to make sure I'm not missing something. In short, if I don't specify an account in the volume request storage class I would expect the destination to be dynamically provisioned the exact same way it would be if there was not a data source specified. I have no expectation for specifying a data *source* to influence the data *destination*.

andyzhangx:
> I have some concerns about these changes, will review more thoroughly tomorrow to make sure I'm not missing something. In short, if I don't specify an account in the volume request storage class I would expect the destination to be dynamically provisioned the exact same way it would be if there was not a data source specified. I have no expectation for specifying a data _source_ to influence the data _destination_.

@acortelyou the original behavior is copying the volume into the same account as source volume, we should keep the same behavior unless user specifies `storageAccount` in storage class. Also if we copy the volume to the account with dynamic matched by the driver, it's difficult for user to find out which destination account has the copied volume since the driver has it's own logic to search a matching account. that's the reason why we only supports copying the volume to the source account in the beginning.

acortelyou:
Thanks @andyzhangx, I appreciate your quick responses and willingness to collaborate on a solution.

I believe the user expects their volume to be provisioned as configured whether or not they provide a `datasource`.



> the original behavior is copying the volume into the same account as source volume

Correct, this is a bug and contradicts the documentation.

> we should keep the same behavior unless user specifies `storageAccount` in storage class

I disagree. We should have the volume provisioning behavior always match the documentation and user intent.

> it's difficult for user to find out which destination account has the copied volume

The documentation provides clear guidance on where/how volumes are to be provisioned and also provides the means to locate and co-locate the volumes if desired.

```
parameters:
  subscriptionID: >
    Specify Azure subscription ID where blob storage directory will be created.
  resourceGroup: >
    If empty, driver will use the same resource group name as current cluster.
  storageAccount: >
    When a specific storage account name is not provided, the driver will look for a suitable storage account that matches the account settings within the same resource group. 
    If it fails to find a matching storage account, it will create a new one. 
    However, if a storage account name is specified, the storage account must already exist.
  containerName: >
    If empty, driver creates a new container name, starting with pvc-fuse for blobfuse or pvc-nfs for NFS v3.
  containerNamePrefix: >
    Specify Azure storage directory prefix created by driver.
  tags: >
    Specify tags for storage account created by the driver.
  matchTags: >
    Match tags when driver tries to find a suitable storage account.
  skuName: >
    Specify skuName for storage account.
  location: >
    Specify location for storage account.
```

https://learn.microsoft.com/en-us/azure/aks/azure-csi-blob-storage-provision#storage-class-parameters-for-dynamic-persistent-volumes

I suspect we'll want to refactor some of these changes so that we can fail faster before potentially creating an account or container, but it would require additional modifications to the tests.

Thanks!

FYI I pulled the logs for the last month and it looks like no-one other than my team and the E2EAKS runners are even trying to use blob volume cloning. I suspect it's because the current behavior is so incredibly limiting. After fixing this bug, blob volume cloning will be extremely useful for a bunch of scenarios. Blobfuse2 can be way more performant than the other csi options we've benchmarked and we want to see it succeed.

pkg/blob/controllerserver.go
```diff
@@ -781,9 +779,9 @@ func (d *Driver) copyBlobContainer(req *csi.CreateVolumeRequest, accountSasToken
 	case util.AzcopyJobRunning:
 		return fmt.Errorf("wait for the existing AzCopy job to complete, current copy percentage is %s%%", percent)
 	case util.AzcopyJobNotFound:
-		klog.V(2).Infof("copy blob container %s to %s", srcContainerName, dstContainerName)
+		klog.V(2).Infof("copy blob container %s to %s", srcPath, dstPath)
 		execFunc := func() error {
-			cmd := exec.Command("azcopy", "copy", srcPath, dstPath, "--recursive", "--check-length=false")
+			cmd := exec.Command("azcopy", "copy", srcPath+accountSasToken, dstPath+accountSasToken, "--recursive", "--check-length=false")
```
andyzhangx, line 784: this is using the same account sas token, if source account and dest account are different, it would break.

andyzhangx: should get srcAccountSASToken if it's two accounts, e.g. 
```
	srcAccountSasToken := dstAccountSasToken
	if srcAccountName != dstAccountName && dstAccountSasToken != "" {
		srcAccountOptions := &azure.AccountOptions{
			Name:                srcAccountName,
			ResourceGroup:       accountOptions.ResourceGroup,
			SubscriptionID:      accountOptions.SubscriptionID,
			GetLatestAccountKey: accountOptions.GetLatestAccountKey,
		}
andyzhangx marked this conversation as resolved.
		if srcAccountSasToken, _, err = d.getAzcopyAuth(ctx, srcAccountName, "", storageEndpointSuffix, srcAccountOptions, nil, "", secretNamespace, true); err != nil {
			return err
		}
	}
```

pkg/blob/controllerserver.go
```diff
@@ -435,19 +442,19 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		setKeyValueInMap(parameters, containerNameField, validContainerName)
 	}
 
+	klog.V(2).Infof("begin to create container(%s) on account(%s) type(%s) subsID(%s) rg(%s) location(%s) size(%d)", validContainerName, accountName, storageAccountType, subsID, resourceGroup, location, requestGiB)
+	if err := d.CreateBlobContainer(ctx, subsID, resourceGroup, accountName, validContainerName, secrets); err != nil {
```
andyzhangx, line 446: azcopy would create container before volume clone, so let's revert this

acortelyou: My intent here was for `CreateBlobContainer()` to configure `Access`, `PublicAccess`, `Metadata` uniformly. I see metadata on azcopy but not public access.

pkg/blob/controllerserver.go
acortelyou: I pushed changes which better encapsulate the volume content source validation and copy, including src sas token and associated tests. Please check that `secrets`, `secretName`, `secretNamespace` logic is correct.

pkg/blob/controllerserver.go
acortelyou: ah, i see that actually an account setting not container setting i will make the change with metadata includes

acortelyou: I have slight reservations still about not creating the blob container first with `CreateBlobContainer()` because azcopy can't default the blob container to private, so there's a slight difference in behavior.

However, since the default for the storage container is private this should only change the outcome when using a static storage account with public access turned on or when explicitly requested.

I left it as a separate commit so that you can decide whether that's a concern and either revert or squash.

acortelyou: Ah, nope, the sanity check test threw this `failed to parse user input due to error: content-type, content-encoding, content-language, content-disposition, cache-control, or metadata is not supported while copying from service to service\n",`. I'll revert the change.

pkg/blob/controllerserver.go
```diff
@@ -367,9 +367,15 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		mc.ObserveOperationWithResult(isOperationSucceeded, VolumeID, volumeID)
 	}()
 
+	secrets := req.GetSecrets()
+
+	volumeContentSourceSASToken, volumeContentSourcePath, err := d.validateVolumeContentSource(ctx, req, storageEndpointSuffix, secrets, secretName, secretNamespace, getLatestAccountKey)
```
umagnus, line 372: I think we don't need to add a function `validateVolumeContentSource` here. We can  move it to `copyBlobContainer` and if `srcAccountName != dstAccountName`, then generate `srcSASToken`, otherwise we just need to generate sas token once.

acortelyou: The intent was to confirm we have have access to the volume source before provisioning a volume we can't populate. I could add a check to skip the sas generation for the *destination* if it matches the source. The added function was just to isolate the changes and make it easier to test, I can fold it back in if you feel strongly about it.

acortelyou: @umagnus on review, the generated sas token is cached. there is no need for an additional optimization.

andyzhangx, line 372: pls only call `validateVolumeContentSource` in the `if req.GetVolumeContentSource() != nil { ..}` logic, that would more clear logic

acortelyou:
@umagnus I have removed the extraneous `funcs` and added a check before regenerating the azcopy auth env.
@andyzhangx I have added `if volContentSource != nil` around all appropriate code blocks and comments to show the logical code flow.

One now-redundant test has been removed.
I have reverted most non-essential changes and scoped this PR tightly to the issue at hand.

acortelyou:
I think I spotted a volume cloning bug where `GetAzcopyJob` was grep'ing using the request `volName` instead of the actual destination `validContainerName`, which could result in multiple duplicate azcopy jobs being started.

I left that fix as a separate commit so that it can be easily reverted if the existing behavior was intentional.

pkg/blob/controllerserver.go
```diff
@@ -67,6 +67,9 @@ const (
 
 // CreateVolume provisions a volume
 func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
+
+	// Request
```
andyzhangx, line 71: pls remove these comments and new empty lines

pkg/blob/controllerserver.go
```diff
@@ -81,19 +84,14 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
-	}
-	defer d.volumeLocks.Release(volName)
-
 	volSizeBytes := int64(req.GetCapacityRange().GetRequiredBytes())
 	requestGiB := int(util.RoundUpGiB(volSizeBytes))
 
+	volContentSource := req.GetVolumeContentSource()
+	secrets := req.GetSecrets()
+
+	// Parameters
```
andyzhangx, line 93: pls remove these comments

pkg/blob/controllerserver.go
```diff
@@ -367,9 +378,53 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		mc.ObserveOperationWithResult(isOperationSucceeded, VolumeID, volumeID)
 	}()
 
+	// Volume Source
```
andyzhangx, line 381: remove

pkg/blob/controllerserver.go
```diff
@@ -367,9 +378,53 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+			if err != nil {
+				return nil, status.Errorf(codes.Internal, "failed to getAzcopyAuth on account(%s) rg(%s), error: %v", accountOptions.Name, accountOptions.ResourceGroup, err)
+			}
+			srcPath = fmt.Sprintf("https://%s.blob.%s/%s", srcAccountName, storageEndpointSuffix, srcContainerName)
+		default:
+			return nil, status.Errorf(codes.InvalidArgument, "VolumeContentSource is not recognized: %v", volContentSource)
+		}
+	}
+
+	// Storage Account
```
andyzhangx, line 424: remove

pkg/blob/controllerserver.go
```diff
@@ -423,33 +478,32 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
-	containerName = replaceWithMap(containerName, containerNameReplaceMap)
-	validContainerName := containerName
-	if validContainerName == "" {
-		validContainerName = volName
-		if containerNamePrefix != "" {
-			validContainerName = containerNamePrefix + "-" + volName
-		}
-		validContainerName = getValidContainerName(validContainerName, protocol)
-		setKeyValueInMap(parameters, containerNameField, validContainerName)
+	// Blob Container
```
andyzhangx, line 481: remove

pkg/blob/controllerserver.go
```diff
@@ -423,33 +478,32 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+	klog.V(2).Infof("begin to create container(%s) on account(%s) type(%s) subsID(%s) rg(%s) location(%s) size(%d)", validContainerName, accountName, storageAccountType, subsID, resourceGroup, location, requestGiB)
+	if err := d.CreateBlobContainer(ctx, subsID, resourceGroup, accountName, validContainerName, secrets); err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to create container(%s) on account(%s) type(%s) rg(%s) location(%s) size(%d), error: %v", validContainerName, accountName, storageAccountType, resourceGroup, location, requestGiB, err)
 	}
 
-	if req.GetVolumeContentSource() != nil {
-		accountSASToken, authAzcopyEnv, err := d.getAzcopyAuth(ctx, accountName, accountKey, storageEndpointSuffix, accountOptions, secrets, secretName, secretNamespace)
-		if err != nil {
-			return nil, status.Errorf(codes.Internal, "failed to getAzcopyAuth on account(%s) rg(%s), error: %v", accountOptions.Name, accountOptions.ResourceGroup, err)
+	// Volume Content
```
andyzhangx, line 488: remove

pkg/blob/controllerserver.go
```diff
@@ -466,6 +520,8 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		}
 	}
 
+	// Response
```
andyzhangx, line 523: remove

pkg/blob/controllerserver.go
```diff
@@ -367,9 +378,53 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+
+	if acquired := d.volumeLocks.TryAcquire(volName); !acquired {
+		// logging the job status if it's volume cloning
+		if volContentSource != nil {
+			jobState, percent, err := d.azcopy.GetAzcopyJob(validContainerName, []string{})
+			return nil, status.Errorf(codes.Aborted, volumeOperationAlreadyExistsWithAzcopyFmt, volName, jobState, percent, err)
+		}
+		return nil, status.Errorf(codes.Aborted, volumeOperationAlreadyExistsFmt, volName)
+	}
+	defer d.volumeLocks.Release(volName)
```
umagnus, line 391: This part can be moved to L362

pkg/blob/controllerserver.go
```diff
@@ -423,31 +467,24 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
 
-	if req.GetVolumeContentSource() != nil {
-		accountSASToken, authAzcopyEnv, err := d.getAzcopyAuth(ctx, accountName, accountKey, storageEndpointSuffix, accountOptions, secrets, secretName, secretNamespace)
-		if err != nil {
-			return nil, status.Errorf(codes.Internal, "failed to getAzcopyAuth on account(%s) rg(%s), error: %v", accountOptions.Name, accountOptions.ResourceGroup, err)
+	if volContentSource != nil {
+		dstAzcopyAuthEnv := srcAzcopyAuthEnv
+		dstAccountSASToken := srcAccountSASToken
+		if srcSubscriptionID != subsID || srcResourceGroupName != resourceGroup || srcAccountName != accountName {
+			if dstAccountSASToken, dstAzcopyAuthEnv, err = d.getAzcopyAuth(ctx, accountName, accountKey, storageEndpointSuffix, accountOptions, secrets, secretName, secretNamespace); err != nil {
```
andyzhangx, line 479: `if volContentSource != nil`, I think `dstAccountSASToken, dstAzcopyAuthEnv, err = d.getAzcopyAuth` could be done before createBlobContainer, that would be better

andyzhangx: make all the auth check before createBlobContainer

andyzhangx (approved):
/lgtm

PR #1517, opened 2024-07-30
fix: volume cloning between different storage account for BlobAccessTierNotSupportedForAccountType error

umagnus:
**What type of PR is this?**

/kind bug

fix: volume cloning between different storage account for BlobAccessTierNotSupportedForAccountType error
azcopy between different storage account has this issue:
image
referring to https://github.com/Azure/azure-storage-azcopy/issues/327 to fix
Also in this issue comment: https://github.com/Azure/azure-storage-azcopy/issues/327#issuecomment-481824121, azcopy may automatically replicate the source blob's tier info to the destination, so create blob container before azcopy instead of using azcopy to create container is more correct.
add volume cloning between different account e2e test

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/blob/controllerserver.go
+++ b/pkg/blob/controllerserver.go
@@ -802,7 +802,7 @@ func (d *Driver) copyBlobContainer(authAzcopyEnv []string, srcPath string, srcAc
 	case util.AzcopyJobNotFound:
 		klog.V(2).Infof("copy blob container %s to %s", srcPath, dstContainerName)
 		execFunc := func() error {
-			cmd := exec.Command("azcopy", "copy", srcPath+srcAccountSASToken, dstPath+dstAccountSASToken, "--recursive", "--check-length=false")
+			cmd := exec.Command("azcopy", "copy", srcPath+srcAccountSASToken, dstPath+dstAccountSASToken, "--recursive", "--check-length=false", "--s2s-preserve-access-tier=false")
 			if len(authAzcopyEnv) > 0 {
 				cmd.Env = append(os.Environ(), authAzcopyEnv...)
 			}
--- a/test/e2e/dynamic_provisioning_test.go
+++ b/test/e2e/dynamic_provisioning_test.go
@@ -1006,4 +1006,79 @@ var _ = ginkgo.Describe("[blob-csi-e2e] Dynamic Provisioning", func() {
 		}
 		test.Run(ctx, cs, ns)
 	})
+
+	ginkgo.It("should clone a volume from an existing NFSv3 volume to another storage class [nfs]", func(ctx ginkgo.SpecContext) {
+		pod := testsuites.PodDetails{
+			Cmd: "echo 'hello world' > /mnt/test-1/data && grep 'hello world' /mnt/test-1/data",
+			Volumes: []testsuites.VolumeDetails{
+				{
+					ClaimSize: "10Gi",
+					MountOptions: []string{
+						"nconnect=8",
+					},
+					VolumeMount: testsuites.VolumeMountDetails{
+						NameGenerate:      "test-volume-",
+						MountPathGenerate: "/mnt/test-",
+					},
+				},
+			},
+		}
+		podWithClonedVolume := testsuites.PodDetails{
+			Cmd: "grep 'hello world' /mnt/test-1/data",
+		}
+		test := testsuites.DynamicallyProvisionedVolumeCloningTest{
+			CSIDriver:           testDriver,
+			Pod:                 pod,
+			PodWithClonedVolume: podWithClonedVolume,
+			StorageClassParameters: map[string]string{
+				"skuName":              "Premium_LRS",
+				"protocol":             "nfs",
+				"mountPermissions":     "0755",
+				"allowsharedkeyaccess": "true",
+			},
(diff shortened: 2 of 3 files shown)


pkg/blob/controllerserver.go
```diff
@@ -366,36 +366,33 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
-				GetLatestAccountKey: getLatestAccountKey,
-			}
-			srcAccountSASToken, srcAzcopyAuthEnv, err = d.getAzcopyAuth(ctx, srcAccountName, "", storageEndpointSuffix, srcAccountOptions, secrets, secretName, secretNamespace)
-			if err != nil {
-				return nil, status.Errorf(codes.Internal, "failed to getAzcopyAuth on account(%s) rg(%s), error: %v", accountOptions.Name, accountOptions.ResourceGroup, err)
-			}
-			srcPath = fmt.Sprintf("https://%s.blob.%s/%s", srcAccountName, storageEndpointSuffix, srcContainerName)
-		default:
-			return nil, status.Errorf(codes.InvalidArgument, "VolumeContentSource is not recognized: %v", volContentSource)
+			// 	var srcVolumeID string
```
andyzhangx, line 375: pls remove commented code if it's not needed any more

pkg/blob/controllerserver.go
```diff
@@ -466,18 +463,25 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to getAzcopyAuth on account(%s) rg(%s), error: %v", accountOptions.Name, accountOptions.ResourceGroup, err)
 		}
-
-		dstPath := fmt.Sprintf("https://%s.blob.%s/%s", accountName, storageEndpointSuffix, validContainerName)
-		if err := d.copyBlobContainer(dstAzcopyAuthEnv, srcPath, srcAccountSASToken, dstPath, dstAccountSASToken, validContainerName); err != nil {
+		if err := d.copyVolume(ctx, req, accountName, accountSASToken, authAzcopyEnv, validContainerName, secretNamespace, accountOptions, storageEndpointSuffix); err != nil {
 			return nil, err
 		}
+		// dstAzcopyAuthEnv := srcAzcopyAuthEnv
```
andyzhangx, line 473: remove

pkg/blob/controllerserver.go
umagnus: done for mistake

pkg/blob/controllerserver.go
umagnus: done

andyzhangx (commented):
/ok-to-test

pkg/blob/controllerserver.go
```diff
@@ -366,36 +366,12 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	defer d.volumeLocks.Release(volName)
 
 	requestName := "controller_create_volume"
-
-	var srcAzcopyAuthEnv []string
-	var srcSubscriptionID, srcResourceGroupName, srcAccountName, srcContainerName, srcPath, srcAccountSASToken string
 	if volContentSource != nil {
 		switch volContentSource.Type.(type) {
 		case *csi.VolumeContentSource_Snapshot:
-			return nil, status.Errorf(codes.InvalidArgument, "VolumeContentSource Snapshot is not yet implemented")
```
andyzhangx, line 375: this should return error, L372 does not make sense

umagnus: `Type` is an `isVolumeContentSource_Type` value and `volContentSource.Type` is not existed when `volContentSource` is nil, so L372 is needed I think. Return error here

pkg/blob/controllerserver.go
```diff
@@ -366,36 +366,12 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
-				ResourceGroup:       srcResourceGroupName,
-				GetLatestAccountKey: getLatestAccountKey,
-			}
-			srcAccountSASToken, srcAzcopyAuthEnv, err = d.getAzcopyAuth(ctx, srcAccountName, "", storageEndpointSuffix, srcAccountOptions, secrets, secretName, secretNamespace)
-			if err != nil {
-				return nil, status.Errorf(codes.Internal, "failed to getAzcopyAuth on account(%s) rg(%s), error: %v", accountOptions.Name, accountOptions.ResourceGroup, err)
-			}
-			srcPath = fmt.Sprintf("https://%s.blob.%s/%s", srcAccountName, storageEndpointSuffix, srcContainerName)
-		default:
-			return nil, status.Errorf(codes.InvalidArgument, "VolumeContentSource is not recognized: %v", volContentSource)
```
andyzhangx, line 398: we still need return error here?

umagnus: we will return error in `copyVolume` function in L829

andyzhangx (approved):
/lgtm

PR #1539, opened 2024-08-05
doc: improve volume cloning doc

umagnus:
**What type of PR is this?**

/kind documentation

doc: improve volume cloning doc

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/deploy/example/cloning/README.md
+++ b/deploy/example/cloning/README.md
@@ -7,6 +7,7 @@
 ## Prerequisites
 - Make sure that the virtual network, where the driver controller pod is deployed, is added to the allowed list of the storage account's VNet setting.
 - Before proceeding, ensure that the application is not writing data to the source volume.
+- In order to allow the use of a managed identity, the source storage account requires the `Storage Blob Data Reader` role, while the destination storage account requires the `Storage Blob Data Contributor` role. If the Storage Blob Data role is not granted, the CSI driver will use a SAS token as a fallback.
 
 ## Create a Source PVC
 


andyzhangx (commented):
In order to allow the use of a managed identity, the source storage account requires the Storage Blob Data Reader role, while the destination storage account requires the Storage Blob Data Contributor role. If the Storage Blob Data role is not granted, the CSI driver will use a SAS token as a fallback.

deploy/example/cloning/README.md
```diff
@@ -7,6 +7,7 @@
 ## Prerequisites
 - Make sure that the virtual network, where the driver controller pod is deployed, is added to the allowed list of the storage account's VNet setting.
 - Before proceeding, ensure that the application is not writing data to the source volume.
+- You can assign `Storage Blob Data Contributor` role to controller identity in storage account to use system-wide/user-assigned managed identity to authorize, otherwise CSI will use SAS token. If you clone a volume between different storage accounts, you can just assign `Storage Blob Data Reader` role in source storage account and `Storage Blob Data Contributor` in destination storage account.
```
andyzhangx, line 10: In order to allow the use of a managed identity, the source storage account requires the Storage Blob Data Reader role, while the destination storage account requires the Storage Blob Data Contributor role. If the Storage Blob Data role is not granted, the CSI driver will use a SAS token as a fallback.

umagnus: done

andyzhangx (approved):
/ok-to-test
/lgtm

PR #1557, opened 2024-08-29
fix: add fallback to sas token on azcopy copy command

umagnus:
**What type of PR is this?**

/kind bug

fix: add fallback to sas token on azcopy copy command
when cx only give `Storage Blob Data Reader` role, azcopy list command will succeed, add fallback to sas token on azcopy copy command when occured AuthorizationPermissionMismatch error

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/blob/controllerserver.go
+++ b/pkg/blob/controllerserver.go
@@ -436,12 +436,22 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		return nil, status.Errorf(codes.Internal, "failed to create container(%s) on account(%s) type(%s) rg(%s) location(%s) size(%d), error: %v", validContainerName, accountName, storageAccountType, resourceGroup, location, requestGiB, err)
 	}
 	if volContentSource != nil {
-		accountSASToken, authAzcopyEnv, err := d.getAzcopyAuth(ctx, accountName, accountKey, storageEndpointSuffix, accountOptions, secrets, secretName, secretNamespace)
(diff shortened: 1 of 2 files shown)


pkg/blob/controllerserver.go
```diff
@@ -802,6 +803,33 @@ func (d *Driver) copyBlobContainer(ctx context.Context, req *csi.CreateVolumeReq
 			return fmt.Errorf("timeout waiting for copy blob container %s to %s complete, current copy percent: %s%%", srcContainerName, dstContainerName, percent)
 		}
 		copyErr := util.WaitUntilTimeout(time.Duration(d.waitForAzCopyTimeoutMinutes)*time.Minute, execFunc, timeoutFunc)
+		if dstAccountSasToken == "" && copyErr != nil && strings.Contains(copyErr.Error(), authorizationPermissionMismatch) {
+			klog.Warningf("azcopy copy failed with AuthorizationPermissionMismatch error, should assign \"Storage Blob Data Contributor\" role to controller identity, fall back to use sas token, original error: %v", copyErr)
+			var srcSasToken, dstSasToken string
```
andyzhangx, line 808: pls use `copyBlobContainer` to avoid duplicate codes, you could call `copyBlobContainer` func twice: 1) use default auth 2) if returns with `authorizationPermissionMismatch` then call `copyBlobContainer` func again to use sas token by force

umagnus: done, use this logic in `copyVolume`

andyzhangx (commented):
/ok-to-test

andyzhangx:
your PR is based on old branch, I cannot patch my modified tests into your PR, so I created another new PR here: https://github.com/kubernetes-sigs/blob-csi-driver/pull/1564

PR #1673, opened 2024-11-05
[WIP]feat: support original workload identity for blobfuse2

cvvz:
**What type of PR is this?**
/kind feature

The previous implementation of supporting workload identity is to use workload identity to retrieve account key for mounting. Since blobfuse2 has supported workload identity authentication, we can use workload identity to directly mount without retrieving account key. 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
feat: support original workload identity for blobfuse2 (without retrieve account key)
```


cvvz:
We need to merge https://github.com/Azure/azure-storage-fuse/pull/1556 first and update the docker file and Agent-baker to install latest blobfuse2

PR #1699, opened 2024-11-14
fix: change mount timeout to 1.5min

umagnus:
**What type of PR is this?**

/kind bug

fix: change mount timeout to 1.5min
Previous set mount timeout in csi as 2 min, which will cause context deadline error but not csi timeout error. Change mount timeout to 1.5 minutes

Details:
the k8s kubelet have 2min timeout for every mount, and we have set 2min timeout inside csi driver(that actually costs 2min+ totally for the whole mount process), and when the outer timeout is triggered by k8s kubelet, it would make another MountDevice call, while at that time, the previous mount is still in progress, thus the second mount would never work until in the inner CSI driver mount timeout.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


andyzhangx (approved):
/lgtm

PR #1798, opened 2025-01-20
feat: support workload identity

cvvz:
**What type of PR is this?**
/kind feature

The previous implementation of supporting workload identity is to use workload identity to retrieve account key for mounting. Since blobfuse2 has supported workload identity authentication, we can use workload identity to directly mount without retrieving account key.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
feat: support original workload identity for blobfuse2 (without retrieve account key)
```

### DIFF
--- a/pkg/blob/blob.go
+++ b/pkg/blob/blob.go
@@ -576,14 +576,14 @@ func (d *Driver) GetAuthEnv(ctx context.Context, volumeID, protocol string, attr
 	if clientID != "" {
 		klog.V(2).Infof("clientID(%s) is specified, use workload identity for blobfuse auth", clientID)
 
-		workload_identity_token, err := parseServiceAccountToken(serviceAccountToken)
+		workloadIdentityToken, err := parseServiceAccountToken(serviceAccountToken)
 		if err != nil {
 			return rgName, accountName, accountKey, containerName, authEnv, err
 		}
 
 		authEnv = append(authEnv, "AZURE_STORAGE_SPN_CLIENT_ID="+clientID)
 		authEnv = append(authEnv, "AZURE_STORAGE_SPN_TENANT_ID="+tenantID)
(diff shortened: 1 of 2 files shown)


acortelyou:
Really glad to see some progress on this feature!

I know it's WIP so perhaps you're already planning to address, but I wanted to mention that the token audience varies by cloud so that will need to be dynamic or configurable.

eg:
```
var azureTokenExchange = {
  AzureCloud:         'api://AzureADTokenExchange'
  AzureUSGovernment:  'api://AzureADTokenExchangeUSGov'
  AzureChinaCloud:    'api://AzureADTokenExchangeChina'
}[environment().name]
```

cvvz:
I've tested with the latest blobfuse2 version v2.4.1 on AKS and checked blobfuse2 and csi driver logs. It worked fine.
image

image
I've removed the WIP label. 
 cc @andyzhangx

pkg/blob/blob.go
```diff
@@ -1136,3 +1145,28 @@ func generateVolumeName(clusterName, pvName string, maxLength int) string {
...
+type serviceAccountToken struct {
+	APIAzureADTokenExchange struct {
+		Token               string    `json:"token"`
+		ExpirationTimestamp time.Time `json:"expirationTimestamp"`
+	} `json:"api://AzureADTokenExchange"`
+}
+
+// parseServiceAccountToken parses the bound service account token from the token passed from NodePublishVolume Request.
+// ref: https://kubernetes-csi.github.io/docs/token-requests.html
+func parseServiceAccountToken(tokenStr string) (string, error) {
```
andyzhangx, line 1160: can you add the ut?

andyzhangx (approved):
/lgtm

pkg/blob/blob.go
andyzhangx: pls add the ut in another PR

PR #1870, opened 2025-03-07
Fix: swapped the wrong ne to an eq when checking if it is a fedora

el-aasi:
**What type of PR is this?**

/kind bug

Currently when using the helm chart with `AzureStackCloud` option it will fail with a duplicate volume mount because of some boolean logic mistake. 

Fixes #

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


andyzhangx:
thanks for the PR, I have worked out another PR to make the logic more clear: https://github.com/kubernetes-sigs/blob-csi-driver/pull/1872

PR #1947, opened 2025-04-16
fix: goroutine leak when timeout

cvvz:
**What type of PR is this?**
/kind bug

fix: goroutine leak when timeout

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
fix: goroutine leak when timeout
```

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -404,7 +404,7 @@ type TimeoutFunc func() (err error)
 // WaitUntilTimeout waits for the exec function to complete or return timeout error
 func WaitUntilTimeout(timeout time.Duration, execFunc ExecFunc, timeoutFunc TimeoutFunc) error {
 	// Create a channel to receive the result of the azcopy exec function
-	done := make(chan bool)
+	done := make(chan bool, 1)
 	var err error
 
 	// Start the azcopy exec function in a goroutine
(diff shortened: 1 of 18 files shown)


cvvz:
Btw, I think k8s lib has implemented `WaitUntilTimeout` why do we implement  this from scratch?

pkg/util/util.go
```diff
@@ -404,7 +404,7 @@ type TimeoutFunc func() (err error)
 // WaitUntilTimeout waits for the exec function to complete or return timeout error
 func WaitUntilTimeout(timeout time.Duration, execFunc ExecFunc, timeoutFunc TimeoutFunc) error {
 	// Create a channel to receive the result of the azcopy exec function
-	done := make(chan bool)
+	done := make(chan bool, 1)
```
andyzhangx, line 407: why set buffer size as 1 would fix the issue?

andyzhangx:
> Btw, for `WaitUntilTimeout`, I think k8s lib has similar implementation, why do we implement this from scratch?

we could use k8s lib if there is a suitable function, any recommendation?

pkg/util/util.go
cvvz: The write operation to a channel in Go is blocking unless there is a receiver waiting for it, this means after `WaitUntilTimeout` exit, the goroutine will remain stuck at `done <- true` forever.
After changed to buffered channel, even if there is no receiver, done <- true can successfully write and continue.

andyzhangx:
> > Btw, for `WaitUntilTimeout`, I think k8s lib has similar implementation, why do we implement this from scratch?
> 
> we could use k8s lib if there is a suitable function, any recommendation?

we should use `wait.PollUntilContextTimeout` func

andyzhangx (approved):
/lgtm

PR #2041, opened 2025-06-16
fix: only use the permission bits for chmod

nearora-msft:
**What type of PR is this?**
/kind bug

If the volumeattributes.mountpermissions are set for an nfs volume, we check if the permissions on the mount targetPath are the same as the expected mountPermissions. When this check is made currently, the targetPath mode bits are masked to use only the permission bits.
We should do the same with the mountPermissions bits to ensure that there isn't an unnecessary mismatch.
After the comparison, if the mount permissions are different, this PR ensures that only mount permissions change with the chmod operation and the rest of the bits like gid remain as is.

Related PR: https://github.com/kubernetes-sigs/azurefile-csi-driver/pull/2600

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/blob/blob.go
+++ b/pkg/blob/blob.go
@@ -1092,9 +1092,11 @@ func chmodIfPermissionMismatch(targetPath string, mode os.FileMode) error {
 		return err
 	}
 	perm := info.Mode() & os.ModePerm
-	if perm != mode {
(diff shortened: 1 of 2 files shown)


andyzhangx (approved):
/lgtm

Issue #2113, opened 2025-08-08
Install fails on openshift


davidkarlsen:
**What happened**:

```shell
k logs csi-blob-node-24flf -c install-blobfuse-proxy
+ INSTALL_BLOBFUSE_PROXY=false
+ DISABLE_UPDATEDB=true
+ SET_MAX_OPEN_FILE_NUM=true
+ SET_READ_AHEAD_SIZE=true
+ MIGRATE_K8S_REPO=false
+ READ_AHEAD_KB=15380
+ KUBELET_PATH=/var/lib/kubelet
+ [ /var/lib/kubelet != /var/lib/kubelet ]
+ HOST_CMD=nsenter --mount=/proc/1/ns/mnt
+ nsenter --mount=/proc/1/ns/mnt cat+ grep ^ID=
 /etc/os-release
+ tr -d "
+ cut -d= -f2
+ DISTRIBUTION=debian
+ nsenter --mount=/proc/1/ns/mnt uname -m
+ ARCH=x86_64
+ echo Linux distribution: debian, Arch: x86_64
Linux distribution: debian, Arch: x86_64
+ . ./blobfuse-proxy/install-proxy.sh
+ set -xe
+ [ debian != ubuntu ]
+ echo skip install blobfuse/blobfuse2 for debian....
skip install blobfuse/blobfuse2 for debian....
+ [ false = true ]
+ [ true = true ]
+ nsenter --mount=/proc/1/ns/mnt sysctl -w fs.file-max=9000000
fs.file-max = 9000000
+ updateDBConfigPath=/host/etc/updatedb.conf
+ [ true = true ]
+ [ -f /host/etc/updatedb.conf ]
+ [ true = true ]
+ echo set read ahead size to 15380KB
set read ahead size to 15380KB
+ which awk
+ AWK_PATH=/usr/bin/awk
+ cat
+ nsenter --mount=/proc/1/ns/mnt udevadm control --reload
Failed to send reload request: No such file or directory
```

**What you expected to happen**:

It should register itself successfully 

**How to reproduce it**:

install on openshift 4.18.x

**Anything else we need to know?**:

Why is blobfuse needed at all if one intends to use the driver with nfs?

**Environment**:

- CSI Driver version:
- Kubernetes version (use `kubectl version`):
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

andyzhangx:
[...]

Issue #1645, opened 2024-10-17
CSI Driver installation failed because of ro /usr


rainerleber:
I want to use the CSI Driver on Gardener Clusters. Unfortunately the /usr filesystem is set to read only (https://github.com/gardenlinux/gardenlinux/blob/f27744b247fb05b858fbfcfa272196a871094168/features/gardener/README.md)

Previous v1.24 everything is working

```
+ INSTALL_BLOBFUSE_PROXY=true
+ DISABLE_UPDATEDB=true
+ SET_MAX_OPEN_FILE_NUM=true
+ SET_READ_AHEAD_SIZE=true
+ READ_AHEAD_KB=15380
+ KUBELET_PATH=/var/lib/kubelet
+ [ /var/lib/kubelet != /var/lib/kubelet ]
+ HOST_CMD=nsenter --mount=/proc/1/ns/mnt
+ nsenter --mount=/proc/1/ns/mnt cat /etc/os-release
+ grep ^ID=
+ cut -d= -f2
+ tr -d "
+ DISTRIBUTION=gardenlinux
+ nsenter --mount=/proc/1/ns/mnt uname -m
+ ARCH=x86_64
+ echo Linux distribution: gardenlinux, Arch: x86_64
+ . ./blobfuse-proxy/install-proxy.sh
+ set -xe
+ [ gardenlinux != ubuntu ]
+ echo skip install blobfuse/blobfuse2 for gardenlinux....
+ updateBlobfuseProxy=true
+ [ -f /host/usr/bin/blobfuse-proxy ]
+ [ true = true ]
+ echo copy blobfuse-proxy....
+ rm -rf /host//var/lib/kubelet/plugins/blob.csi.azure.com/blobfuse-proxy.sock
Linux distribution: gardenlinux, Arch: x86_64
skip install blobfuse/blobfuse2 for gardenlinux....
copy blobfuse-proxy....
+ cp /blobfuse-proxy/blobfuse-proxy /host/usr/bin/blobfuse-proxy --force
cp: cannot create regular file '/host/usr/bin/blobfuse-proxy': Read-only file system
```

**What happened**:
CSI Driver will not start

**What you expected to happen**:
CSI Driver is started

**How to reproduce it**:
Set /usr to ro

**Anything else we need to know?**:

**Environment**:

- CSI Driver version: 1.25.0
- Kubernetes version (use `kubectl version`): 1.29.2
[...]

PR #2125, opened 2025-08-09
fix: support blobfuse2 install for GKE cos node

### LINKED ISSUE #2109: install-blobfuse-proxy init-container fails on GKE/COS nodes due to read-only filesystem
The Azure Blob CSI driver (v1.26.3) controller pods deploy successfully, but node pods fail because the install-blobfuse-proxy init-container cannot write to /host/usr/bin on Container-Optimized OS (COS) nodes. 

This is identical to [Issue #1645](https://github.com/kubernetes-sigs/blob-csi-driver/issues/1645) (fixed for Linux Gartner), but COS requires a separate fix due to its immutable filesystem.

Initially it gives error for `/opt` which i fixed by changing it to `/var` but now its giving below error

**Error:**

```
+ INSTALL_BLOBFUSE_PROXY=true
+ DISABLE_UPDATEDB=true
+ SET_MAX_OPEN_FILE_NUM=true
+ SET_READ_AHEAD_SIZE=true
+ MIGRATE_K8S_REPO=false
+ READ_AHEAD_KB=15380
+ KUBELET_PATH=/var/lib/kubelet
+ [ /var/lib/kubelet != /var/lib/kubelet ]
+ HOST_CMD=nsenter --mount=/proc/1/ns/mnt
+ nsenter --mount=/proc/1/ns/mnt cat /etc/os-release
+ grep ^ID=
+ cut -d= -f2
+ tr -d "
+ DISTRIBUTION=cos
+ nsenter --mount=/proc/1/ns/mnt uname -m
+ ARCH=x86_64
+ echo Linux distribution: cos, Arch: x86_64
+ . ./blobfuse-proxy/install-proxy.sh
+ set -xe
Linux distribution: cos, Arch: x86_64
+ [ cos != ubuntu ]
+ echo skip install blobfuse/blobfuse2 for cos....
skip install blobfuse/blobfuse2 for cos....
+ [ true = true ]
+ updateBlobfuseProxy=true
+ [ -f /host/usr/bin/blobfuse-proxy ]
+ [ true = true ]
+ echo copy blobfuse-proxy....
+ rm -rf /host//var/lib/kubelet/plugins/blob.csi.azure.com/blobfuse-proxy.sock
copy blobfuse-proxy....
+ cp /blobfuse-proxy/blobfuse-proxy /host/usr/bin/blobfuse-proxy --force
cp: cannot create regular file '/host/usr/bin/blobfuse-proxy': Read-only file system
```

**Affected Components:**
Working: csi-blob-controller pods (deploy successfully).

Failing: csi-blob-node pods (stuck in Init:Error due to init-container).

**Expected Behavior:**
The init-container should:
Use COS-approved writable paths (e.g., /home/kubernetes/bin).

**Proposed Fix:**
Extend the Linux Gartner fix to:

Add COS detection in the init-container script.

Redirect installation to /home/kubernetes/bin (COS allows writes here).

@andyzhangx whats plan for new release with OS like gardenlinux and cos? It will help to use driver with these kind of OS.

andyzhangx:
**What type of PR is this?**
/kind bug

fix: support blobfuse2 install for GKE cos node

on GKE cos node, we should set following mount option for blobfuse volume due to error `failed to create default work dir [mkdir /.blobfuse2: read-only file system]`:
```
    - --default-working-dir=/home/mnt
    - --tmp-path=/home/mnt
```

Fixes #2109, #1645, #2113

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/blobfuse-proxy/install-proxy-rhcos.sh
+++ b/pkg/blobfuse-proxy/install-proxy-rhcos.sh
@@ -18,8 +18,18 @@ set -xe
 
 BIN_PATH=${BIN_PATH:-/usr/local/bin}
 if [ "${DISTRIBUTION}" = "cos" ]; then
+  echo "set BIN_PATH to /home/kubernetes/bin"
   BIN_PATH="/home/kubernetes/bin"
 fi
+if [ "${DISTRIBUTION}" = "gardenlinux" ]; then
+  echo "set BIN_PATH to /var/bin"
+  BIN_PATH="/var/bin"
+fi
+if [ "${CUSTOM_BIN_PATH}" != "" ]; then
+  echo "set BIN_PATH to ${CUSTOM_BIN_PATH}"
+  BIN_PATH="${CUSTOM_BIN_PATH}"
+fi
+
 #copy blobfuse2 binary to BIN_PATH/blobfuse2
 updateBlobfuse2="true"
 if [ "${INSTALL_BLOBFUSE}" = "true" ] || [ "${INSTALL_BLOBFUSE2}" = "true" ]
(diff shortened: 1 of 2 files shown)


imranrazakhan:
@andyzhangx Can we merge it to master?

pkg/blobfuse-proxy/install-proxy-rhcos.sh
```diff
@@ -34,15 +48,15 @@ else
...
+  chmod 755 /host/${BIN_PATH}/blobfuse2
 fi
 
 if [ "${INSTALL_BLOBFUSE_PROXY}" = "true" ];then
   # install blobfuse-proxy
   updateBlobfuseProxy="true"
-  if [ -f "/host/usr/local/bin/blobfuse-proxy" ];then
-    old=$(sha256sum /host/usr/local/bin/blobfuse-proxy | awk '{print $1}')
+  if [ -f "/host/${BIN_PATH}/blobfuse-proxy" ];then
+    old=$(sha256sum /host/${BIN_PATH}/blobfuse-proxy | awk '{print $1}')
```
the-veloper, line 59: bin path already contains `/`
this causes double `/` like: `/host//home/kubernetes/bin/...`

andyzhangx: that's ok, it should work on Linux, right?

the-veloper: yeah, i tested it and that works, although it is ugly.
problem is that /host is not mounted on the init container, so none of the custom paths can work:
https://github.com/kubernetes-sigs/blob-csi-driver/blob/master/charts/latest/blob-csi-driver/templates/csi-blob-node.yaml#L110

the-veloper: we can mount /home/kubernetes/bin and /var/bin for cos and gardenlinux and introduce additional mounts for when a custom bin path is required, so that we don't expose the entire /host? wdyt @andyzhangx

PR #2137, opened 2025-08-26
feat: support blobfuse2 preview package install

jainakanksha-msft:


pkg/blobfuse-proxy/install-proxy.sh
```diff
@@ -73,7 +73,12 @@ then
       echo "install blobfuse2 with latest version"
       pkg_list="${pkg_list} blobfuse2"
     else
-      pkg_list="${pkg_list} blobfuse2=${BLOBFUSE2_VERSION}"
+      if echo "${BLOBFUSE2_VERSION}" | grep -q "preview"; then
+        $HOST_CMD apt remove -y blobfuse2
```
andyzhangx: not necessary to remove blobfuse2 package here since it's not installed yet

jainakanksha-msft: image
I am experiencing a failure due to a blobfuse package conflict.

andyzhangx:
@jainakanksha-msft can you sign the easycla? select individual contributor

pkg/blobfuse-proxy/install-proxy.sh
andyzhangx: this would return error if blobfuse2 package is not install on the node

jainakanksha-msft: Handled it with a check beforehand

pkg/blobfuse-proxy/install-proxy.sh
```diff
@@ -74,9 +74,14 @@ then
       pkg_list="${pkg_list} blobfuse2"
     else
       if echo "${BLOBFUSE2_VERSION}" | grep -q "preview"; then
-        $HOST_CMD apt remove -y blobfuse2
+        if dpkg -l | grep -q blobfuse2; then
```
andyzhangx: need to run `dpkg -l` comand on the host, another option is `$HOST_CMD apt remove -y blobfuse2 || true`

jainakanksha-msft: Handled

PR #2173, opened 2025-09-12
fix(proxy_installer): add necessary mounts for cos,gardenlinux bin paths

the-veloper:
**What type of PR is this?**

/kind bug

It adds the necessary mounts so that the rhcos installer script can copy the binary. Also adds the option to do additional mounts for custom bin_paths 

Related to: https://github.com/kubernetes-sigs/blob-csi-driver/pull/2125#pullrequestreview-3215803052

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


charts/latest/blob-csi-driver/templates/csi-blob-node.yaml
```diff
@@ -107,12 +107,19 @@ spec:
             - name: ALLOW_PACKAGE_INSTALL_DOWNGRADE
               value: "{{ .Values.node.blobfuseProxy.allowPackageInstallDowngrade }}"
           volumeMounts:
+            - name: host-home-kubernetes-bin
+              mountPath: /host/home/kubernetes/bin
```
andyzhangx, line 112: this volumeMount should not be added by default since it's only for specific os, is it possible to add a helm chart config for these volumeMounts?

the-veloper: do you mean like mounting these based on a value? 

```
linux:
  kubelet: /var/lib/kubelet
  distro: debian
  # Distribution-specific mount flags
  # Only one of these can be true at a time
  mountHomeK8sBin: false  # For distribution COS (Container-Optimized OS)
  mountVarLib: false      # For distribution Garden Linux
  ```
  
  could do something like this in the values? or based on the distro value?

edit: decided to go with the distro value, as it was implemented for `fedora`

Comment (andyzhangx) [reply charts/latest/blob-csi-driver/templates/csi-blob-node.yaml:112]: you already have `additionalVolumeMounts`, so you could set these two mountPath there

Comment (andyzhangx) [reply charts/latest/blob-csi-driver/templates/csi-blob-node.yaml:112]: of use a value to tell whether need to set there

File: charts/latest/blob-csi-driver/templates/csi-blob-node.yaml [hunk 2]
```diff
@@ -284,6 +291,12 @@ spec:
               name: mountpoint-dir
 {{- end }}
       volumes:
+        - name: host-home-kubernetes-bin
```
andyzhangx, line 299: same here

the-veloper: applied conditional mounting based on the `.Values.linux.distro`.
`additionalVolumeMounts` was intended for the custom bin path (outside cos/gardenlinux) or other non mandatory mounts to the pod and would get overwritten if someone passes a value there. wouldnt that be a problem?

charts/latest/blob-csi-driver/templates/csi-blob-node.yaml
the-veloper: used the `linux.distro` to determine which mount to set

charts/latest/blob-csi-driver/templates/csi-blob-node.yaml
```diff
@@ -107,12 +107,23 @@ spec:
...
+{{- end }}
             - name: host-usr
               mountPath: /host/usr
             - name: host-usr-local
               mountPath: /host/usr/local
             - name: host-etc
               mountPath: /host/etc
+{{- if .Values.node.additionalVolumeMounts }}
+{{- toYaml .Values.node.additionalVolumeMounts | nindent 12 }}
+{{- end }}
```
the-veloper, line 126: should i add this to the rest of the containers?

andyzhangx: it's not necessary now since blobfuse-proxy would run on host, it's only needed in init-container

andyzhangx:
latest chart config has changed, pls run "helm package charts/latest/blob-csi-driver -d charts/latest/" to update tgz file

the-veloper:
> latest chart config has changed, pls run "helm package charts/latest/blob-csi-driver -d charts/latest/" to update tgz file

done, does it make sense to add it as a github action? i can do that, if you confirm.

charts/latest/blob-csi-driver/templates/csi-blob-node.yaml
```diff
@@ -284,6 +295,16 @@ spec:
...
       volumes:
+{{- if eq .Values.linux.distro "cos" }}
+        - name: host-home-kubernetes-bin
+          hostPath:
+            path: /home/kubernetes/bin
+{{- end }}
+{{- if eq .Values.linux.distro "gardenlinux" }}
+        - name: host-var-bin
+          hostPath:
+            path: /var/bin
```
andyzhangx, line 306: how do you know it's `/var/bin` for `gardenlinux`?

the-veloper: the script sets it and i have also tested it.
https://github.com/kubernetes-sigs/blob-csi-driver/blob/master/pkg/blobfuse-proxy/install-proxy-rhcos.sh#L26

andyzhangx (approved):
/lgtm

the-veloper:
should this be cherry-picked in the release-1.27 branch also? @andyzhangx

PR #2209, opened 2025-10-17
🌈 Feature: Add support for Flatcar Container Linux

t-lo:
**What type of PR is this?**

/kind bug
/kind feature

This change adds support for [Flatcar](https://www.flatcar.org/) to the blobfuse installer. Changes are minimal; we're re-using the rhcos installer script and the gardenlinux paths.

Changes are minimal; we're re-using the rhcos installer script and the gardenlinux paths.

As Flatcar currently does not ship libfuse, we install it alongside blobfuse2, similar to what https://github.com/kubernetes-sigs/blob-csi-driver/pull/2208 does for RHCOS.

Lastly, this change contains a fix for updating `PATH` in the `blobfuse-proxy.service` systemd service unit. Previously, the way `PATH` was modified led to the variable contents being overwritten instead of added to. See 0f3078073733cd44c55f3c034d62923150a7004d for details.

**Special notes for your reviewer:**

This change was verified by a test cluster deployment by @invidian . Pre-built container images for testing are available from https://github.com/flatcar-hub/blob-csi-driver/pkgs/container/blob-csi .

**Release note**:
```
* Installer: Added support for Flatcar Container Linux
```


andyzhangx:
@t-lo could you also cherrypick this PR to release-1.33? thx

t-lo:
@andyzhangx I'd be happy to help! Could you please elaborate on ` release-1.33` ? I see `release-`  branches for ` 1.23` to `1.27` but no `1.33`?

t-lo:
@andyzhangx I was browsing closed PRs and saw that everything gets backported to 1.26 and 1.27, so I did the same:
* #2213 
* #2214 

Is this what you're looking for?

andyzhangx:
@t-lo `/var/bin/` actually does not exist on flatcar distro, then how to install a systemd service on flatcar distro?

```
+ echo copy blobfuse-proxy....
+ rm -rf /host//var/lib/kubelet/plugins/blob.csi.azure.com/blobfuse-proxy.sock
Linux distribution: flatcar, Arch: x86_64
set BIN_PATH to /var/bin
no need to install blobfuse2
copy blobfuse-proxy....
+ cp /blobfuse-proxy/blobfuse-proxy /host/var/bin/blobfuse-proxy --force
cp: cannot create regular file '/host/var/bin/blobfuse-proxy': No such file or directory
```

t-lo:
@andyzhangx  IIRC `/var/bin` is created in the blobfuse proxy deployment YAML file (the one that also takes care of mounting the host FS into the container's `/host/` directory); @invidian would know more.

PR #2218, opened 2025-10-23
[release-1.26]🐞 fix: shasum in blobfuse-proxy installer for Flatcar, fix PATH in blobfuse-proxy.service

t-lo:
The blobfuse-proxy installer script used with Flatcar, install-proxy-rhcos.sh, uses sha256sum to check whether blobfuse binaries need to be (re-)installed. If the checksum of blobfuse2 on the host and blobfuse2 shipped with the container differ, the host blobfuse2 is replaced.

However, for Flatcar, we use a custom name for the blobfuse2 binary on the host ("blobfuse2.bin") and ship a wrapper script named "blobfuse2" to update the linker path for blobfuse2. The shasum check did not account for this and inadvertently generated a checksum for the wrapper script instead of the actual binary.

Additionally, it fixes an issue introduced with the original Flatcar support PR in https://github.com/kubernetes-sigs/blob-csi-driver/pull/2209/commits/0f3078073733cd44c55f3c034d62923150a7004d, which modifies the PATH environment variable correctly for `blobfuse-proxy/blobfuse-proxy.service` but did not actually commit those changes to the service unit file.

This PR fixes both.

Backport of https://github.com/kubernetes-sigs/blob-csi-driver/pull/2216 to release-1.26.

/kind bug


andyzhangx:
addressed by https://github.com/kubernetes-sigs/blob-csi-driver/pull/2222

PR #2248, opened 2025-11-13
feat: add telemetry as part of mount options for blobfuse mounts

mittachaitu:
**What type of PR is this?**
/ kind bug

This PR adds telemetry=blobpartner-csi/blob-driver-version as mount options for blobfuse mounts.
This is required to track, usage of Azure Blob Storage through AKS Blob CSI driver.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

- Need to check whether this tag is reflecting at blob storage metrics.

**Release note**:
```
none
```

### DIFF
--- a/pkg/blobfuse-proxy/server/server.go
+++ b/pkg/blobfuse-proxy/server/server.go
@@ -34,8 +34,12 @@ import (
 )
 
 var (
-	mutex sync.Mutex
+	mutex         sync.Mutex
+	driverVersion string
 )
+
+// telemetryTagPrefix is used to identify the mounts done via blobcsi driver
+const telemetryTagPrefix = "blobpartner-csi/"
 
 type BlobfuseVersion int
 
@@ -47,12 +51,14 @@ const (
 type MountServer struct {
 	blobfuseVersion BlobfuseVersion
 	mount_azure_blob.UnimplementedMountServiceServer
+	exec func(name string, arg ...string) *exec.Cmd
 }
 
 // NewMountServer returns a new Mountserver
 func NewMountServiceServer() *MountServer {
 	mountServer := &MountServer{}
 	mountServer.blobfuseVersion = getBlobfuseVersion()
+	mountServer.exec = exec.Command
 	return mountServer
 }
 
@@ -82,13 +88,15 @@ func (server *MountServer) MountAzureBlob(_ context.Context,
 			klog.V(2).Infof("append --disable-version-check to mount args")
 			args = args + " " + "--disable-version-check=true"
 		}
+		// Adding telemetry tag to know that blob is been mounted through AKS CSI Driver
+		args = addTelemetryTagToArgs(args)
 		args = util.TrimDuplicatedSpace(args)
 		klog.V(2).Infof("mount with v2, protocol: %s, args: %s", protocol, args)
-		cmd = exec.Command("blobfuse2", strings.Split(args, " ")...)
+		cmd = server.exec("blobfuse2", strings.Split(args, " ")...)
 	} else {
 		args = util.TrimDuplicatedSpace(args)
 		klog.V(2).Infof("mount with v1, protocol: %s, args: %s", protocol, args)
-		cmd = exec.Command("blobfuse", strings.Split(args, " ")...)
+		cmd = server.exec("blobfuse", strings.Split(args, " ")...)
 	}
 
 	cmd.Env = append(os.Environ(), authEnv...)
@@ -152,4 +160,38 @@ func getBlobfuseVersion() BlobfuseVersion {
 
 	klog.V(2).Info("proxy default using blobfuse V1 for mounting")
 	return BlobfuseV1
+}
+
+// Adding CSI driver specific telemetry tag if not present to
+// know that blob is been mounted through CSI Driver
+func addTelemetryTagToArgs(args string) string {
+	telemetryTag := telemetryTagPrefix + driverVersion
+	if !strings.Contains(args, "--telemetry") {
+		klog.V(2).Infof("append --telemetry=%s to mount args", telemetryTag)
+		args = args + " " + "--telemetry=" + telemetryTag
+	} else {
+		// If telemetry flag is already present, check for aks tag if not present
+		// then user might have their own telemetry tag append aks tag to it
+		if !strings.Contains(args, telemetryTagPrefix) {
+			argSlice := strings.Fields(args)
+			telemetryIndex := -1
+			for i, arg := range argSlice {
+				if strings.HasPrefix(arg, "--telemetry=") {
+					telemetryIndex = i
+					break
+				}
+			}
+			if telemetryIndex != -1 {
+				// Update the telemetry tag, appending our tag if not present
+				telemetryVal := strings.TrimPrefix(argSlice[telemetryIndex], "--telemetry=")
+				// Avoid duplicating the tag if already present
+				if !strings.Contains(telemetryVal, telemetryTag) {
+					argSlice[telemetryIndex] = "--telemetry=" + telemetryTag + "," + telemetryVal
+					args = strings.Join(argSlice, " ")
+					klog.V(2).Infof("updated --telemetry tag in mount args: %s", args)
+				}
+			}
+		}
+	}
+	return args
 }
--- a/pkg/blobfuse-proxy/server/server_test.go
+++ b/pkg/blobfuse-proxy/server/server_test.go
@@ -18,6 +18,8 @@ package server
 
 import (
 	"context"
+	"os/exec"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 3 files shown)


pkg/blob/nodeserver.go
```diff
@@ -409,6 +409,7 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 	}
 
 	// Get mountOptions that the volume will be formatted and mounted with
+	mountFlags = util.JoinMountOptions(mountFlags, []string{fmt.Sprintf("--telemetry=azpartner-aks/%s", d.Version)})
```
andyzhangx: since blobfuse v1 does not support this option, it would break, pls add telemetry into blobfuse2 example config: https://github.com/kubernetes-sigs/blob-csi-driver/blob/master/deploy/example/storageclass-blobfuse2.yaml

mittachaitu: This telemetry is to track usage of azure blob storage through AKS, we can't add it in example config since there will be high chances that user can avoid it. I have updated mountFlag only if protocol is fuse

pkg/blob/nodeserver.go
```diff
@@ -409,6 +409,10 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 	}
 
 	// Get mountOptions that the volume will be formatted and mounted with
+	if protocol == Fuse2 {
+		// Adding telemetry tag to know that blob is been mounted through AKS
+		mountFlags = util.JoinMountOptions(mountFlags, []string{fmt.Sprintf("--telemetry=azpartner-aks/%s", d.Version)})
```
andyzhangx: follow the PR example here: https://github.com/kubernetes-sigs/blob-csi-driver/pull/2024 since protocol may be empty here, and also check whether this option is already specified

andyzhangx: btw, I think just use `--telemetry=aks/%s` is more straightfoward

pkg/blobfuse-proxy/server/server.go
```diff
@@ -82,6 +84,21 @@ func (server *MountServer) MountAzureBlob(_ context.Context,
 			klog.V(2).Infof("append --disable-version-check to mount args")
 			args = args + " " + "--disable-version-check=true"
 		}
+		// Adding telemetry tag to know that blob is been mounted through AKS CSI Driver
+		if !strings.Contains(args, "--telemetry") {
+			klog.V(2).Infof("append --telemetry=%s to mount args", telemetryTag)
+			args = args + " " + "--telemetry=" + telemetryTag
+		} else {
+			// If telemetry flag is already present, check for aks tag if not present
```
andyzhangx: If telemetry flag is already present, we just respect original setting, I don't think we need to change the original value by force

pkg/blob/nodeserver.go
mittachaitu: Andy, I have moved the code to under pkg/blobfuse-proxy/server.go and if user specifies the option through StorageClass then we are appending above to telemetry option

ex: In below example user has provided telemetry as part of mount option... 
```yaml
mountOptions:
  - -o allow_other
  - --file-cache-timeout-in-seconds=120
  - --use-attr-cache=true
  - --cancel-list-on-mount-seconds=10  # prevent billing charges on mounting
  - -o attr_timeout=120
  - -o entry_timeout=120
  - -o negative_timeout=120
  - --telemetry=tire1
```

Then mount args to blobfuse2 will be (--telemetry=azpartner-aks/v1.28.0,normaltest)
```
 server.go:99] updated --telemetry tag in mount args: mount /var/lib/kubelet/plugins/kubernetes.io/csi/blob.csi.azure.com/d6b00230de0d40736dda3f197b49a4919bfb9b7512f4466d881183c6a2baa974/globalmount -o allow_other --file-cache-timeout-in-seconds=120 --use-attr-cache=true --cancel-list-on-mount-seconds=10 -o attr_timeout=120 -o entry_timeout=120 -o negative_timeout=120 --log-level=LOG_DEBUG --cache-size-mb=1000  --telemetry=azpartner-aks/v1.28.0,normaltest --pre-mount-validate=true --use-https=true --empty-dir-check=false --tmp-path=/mnt/<RG>#<Storage Account>#<Contianer>#<Blob>#default# --container-name=traineddata --ignore-open-flags=true --disable-version-check=true
```

Note: If there is no `--telemetry` in mount option then `--telemetry=azpartner-aks/v1.28.0`. 

> btw, I think just use --telemetry=aks/%s is more straightfoward

The tag has been suggested by @vibhansa-msft, so not sure if there is any monitoring from blobfuse side

pkg/blobfuse-proxy/server/server.go
mittachaitu: If user has some custom tag, then we can't use that for tracking blob usage through AKS. If we just use the original setting then whole purpose to track usage will be lost, so I appended our tag with user specified value.

pkg/blobfuse-proxy/server/server.go
```diff
@@ -82,6 +84,21 @@ func (server *MountServer) MountAzureBlob(_ context.Context,
...
+			// then user might have their own telemetry tag append aks tag to it
+			if !strings.Contains(args, "azpartner-aks") {
+				splitedArgs := strings.Split(args, "--telemetry=")
+				if len(splitedArgs) == 2 {
+					args = splitedArgs[0] + " --telemetry=" + telemetryTag + "," + splitedArgs[1]
```
andyzhangx: could you add some test for this?

pkg/blobfuse-proxy/server/server.go
andyzhangx: but what if customer only wants to keep the original telemetry? this PR would change the telemetry even user does not want to.

and this adds the telemetry as `azpartner-aks` in open source driver, which means if it runs on AWS k8s node, it would still have this `azpartner-aks` telemetry, so maybe use default userAgent as `fmt.Sprintf("%s/%s", driverName, driverVersion)` is more reasonable.

mittachaitu: I have updated tag to `blobpartner-csi/<driver-version>`, Concern from blobfuse team is if we go with driver specific tag then it will become costly operation for blobfuse analytics team to parse these tags (since they are expecting all partners should use similar tag instead of each partner goes with different tag).

pkg/blobfuse-proxy/server/server.go
mittachaitu: Done, Added Unit test

pkg/blobfuse-proxy/server/server.go
```diff
@@ -34,9 +34,13 @@ import (
 )
 
 var (
-	mutex sync.Mutex
+	mutex         sync.Mutex
+	driverVersion string
 )
 
+// telemetryTagPrefix is used to identify the mounts done via blobcsi driver
+const telemetryTagPrefix = "blobpartner-csi/"
```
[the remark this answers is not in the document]
mittachaitu: Updated PR description

pkg/blobfuse-proxy/server/server.go
```diff
@@ -82,13 +89,28 @@ func (server *MountServer) MountAzureBlob(_ context.Context,
...
+		if !strings.Contains(args, "--telemetry") {
+			klog.V(2).Infof("append --telemetry=%s to mount args", telemetryTag)
+			args = args + " " + "--telemetry=" + telemetryTag
+		} else {
+			// If telemetry flag is already present, check for aks tag if not present
+			// then user might have their own telemetry tag append aks tag to it
+			if !strings.Contains(args, telemetryTagPrefix) {
+				splitedArgs := strings.Split(args, "--telemetry=")
+				if len(splitedArgs) == 2 {
+					args = splitedArgs[0] + " --telemetry=" + telemetryTag + "," + splitedArgs[1]
```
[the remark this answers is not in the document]
mittachaitu: We are clearing double space issue later part of code

pkg/blobfuse-proxy/server/server_test.go
```diff
@@ -64,3 +66,95 @@ func TestServerMountAzureBlob(t *testing.T) {
...
+			mountServer: MountServer{
+				blobfuseVersion: BlobfuseV1,
+				exec:            fakeExecCommandEchoArgs,
+			},
+			code: codes.OK,
+			areValidTelemetryArgs: func(cmdArgs string) bool {
+				// No telemetry arg should be added for blobfuse v1
+				return !strings.Contains(cmdArgs, "--telemetry=") && cmdArgs == "--account-name=testaccount --container-name=testcontainer --tmp-path=/tmp/blobfuse-tmp"
+			},
+		},
```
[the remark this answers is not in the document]
mittachaitu: Added test case!!

pkg/blobfuse-proxy/server/server.go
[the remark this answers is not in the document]
mittachaitu: Expectation is that user will provide valid input!!

pkg/blobfuse-proxy/server/server.go
```diff
@@ -82,13 +89,41 @@ func (server *MountServer) MountAzureBlob(_ context.Context,
 			klog.V(2).Infof("append --disable-version-check to mount args")
 			args = args + " " + "--disable-version-check=true"
 		}
+		// Adding telemetry tag to know that blob is been mounted through AKS CSI Driver
+		if !strings.Contains(args, "--telemetry") {
```
andyzhangx: can you wrap L92-L119 as a new function, that would be easier to read and write ut. And pls squash all commits into one when you have the PR ready.

mittachaitu: Moved to a different function and squashed all commits

PR #2249, opened 2025-11-16
fix: snapshot restore and volume clone failure in some new sovereign …

hasethuraman:
**What type of PR is this?**

/kind bug

We use azcopy during volume clone. Azcopy trusts known Microsoft's endpoint suffixes. When a new endpoint is added, AzCopy fails saying it is not trustable & so volume provisioning fails. Since Azure File CSI Driver knows the trusted new endpoint through the config & in such case if the new endpoint is not in azcopy known list, we will use the option that azcopy provides to pass new suffixes

Without this PR, we need to release new version until which the volume clone is not possible

Fixes #

**Requirements**:
  
- [x] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/blob/blob.go
+++ b/pkg/blob/blob.go
@@ -268,6 +268,9 @@ type Driver struct {
 	waitForAzCopyTimeoutMinutes int
 	// azcopy for provide exec mock for ut
 	azcopy *util.Azcopy
+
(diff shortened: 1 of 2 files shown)


andyzhangx (approved):
/lgtm

PR #2305, opened 2026-01-21
feat: support reading service account tokens from CSI secrets field for Kubernetes 1.35+

aramase:
/kind feature

implementing changes for KEP: https://github.com/kubernetes/enhancements/issues/5538

see https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/5538-csi-sa-tokens-secrets-field/README.md#driver-migration-example for why we're doing this.

```release-note
Support reading service account tokens from CSI secrets field for Kubernetes 1.35+
```


aramase:
@andyzhangx what's the best way to handle this in the helm charts? do you create a new version of chart for every Kubernetes release?

for 1.35+, we should have `serviceAccountTokenInSecrets: true` in the `CSIDriver`. See the [KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/5538-csi-sa-tokens-secrets-field) and example for how I'm handling this in Secrets Store CSI Driver: https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/1979 for reference.

landreasyan:
Should we enable the feature in this PR by setting `serviceAccountTokenInSecrets: true` for the latest `CSIDriver` in the charts?

aramase:
> Should we enable the feature in this PR by setting `serviceAccountTokenInSecrets: true` for the latest `CSIDriver` in the charts?

we should!
ref: https://github.com/kubernetes-sigs/blob-csi-driver/pull/2305#issuecomment-3781379134

landreasyan:
> > Should we enable the feature in this PR by setting `serviceAccountTokenInSecrets: true` for the latest `CSIDriver` in the charts?
> 
> we should! ref: [#2305 (comment)](https://github.com/kubernetes-sigs/blob-csi-driver/pull/2305#issuecomment-3781379134)

@aramase, I discussed with @andyzhangx - you can add the `serviceAccountTokenInSecrets: true` to `CSIDriver` in the latest charts in this PR and we will create a new tag, 1.36, once these changes merge! :)

andyzhangx (approved):
/lgtm

I have made some changes to disable `serviceAccountTokenInSecrets: true` (enable in e2e test by default) in `CSIdriver` by default, so this version could still run on k8s cluster version < 1.35. We need to keep backward compatibility in the upstream version since many users are still running older k8s versions.

PR #2326, opened 2026-02-06
chore: add metrics logs back

hccheng72:
**What type of PR is this?**
kind feature

add csi operation latency logs back for existing kusto query.
log example in this PR:
`metrics.go:159] "Observed Request Latency" logger="logLatency" latency_seconds=25.889855108 request="blob_csi_driver_controller_create_volume" success="true" volumeid="capz-51clrs#fuseb801c549aea44f278eb#container-pvc-f94rk#pvc-0a021d1c-4a58-44a7-8993-910cb89444b0#blob-7976#" resource_group="capz-51clrs" subscription_id="46678f10-4bbb-447e-98e8-d2829589f2d8" source="blob.csi.azure.com"`

https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/3464#issuecomment-3782111673
Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/blob/controllerserver.go
+++ b/pkg/blob/controllerserver.go
@@ -398,8 +398,7 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	csiMC := csiMetrics.NewCSIMetricContext(requestName).WithBasicVolumeInfo(d.cloud.ResourceGroup, d.cloud.SubscriptionID, d.Name)
 	isOperationSucceeded := false
 	defer func() {
-		csiMC.WithAdditionalVolumeInfo(VolumeID, volumeID)
-		csiMC.ObserveWithLabels(isOperationSucceeded, "storage_account_type", storageAccountType, "protocol", protocol)
+		csiMC.WithAdditionalVolumeInfo(VolumeID, volumeID).ObserveWithLabels(isOperationSucceeded, csiMetrics.StorageAccountType, storageAccountType, csiMetrics.Protocol, protocol)
 	}()
 
 	var accountKey string
@@ -566,8 +565,7 @@ func (d *Driver) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest)
 	csiMC := csiMetrics.NewCSIMetricContext("controller_delete_volume").WithBasicVolumeInfo(d.cloud.ResourceGroup, d.cloud.SubscriptionID, d.Name)
 	isOperationSucceeded := false
 	defer func() {
-		csiMC.WithAdditionalVolumeInfo(VolumeID, volumeID)
-		csiMC.Observe(isOperationSucceeded)
+		csiMC.WithAdditionalVolumeInfo(VolumeID, volumeID).Observe(isOperationSucceeded)
 	}()
 
 	if resourceGroupName == "" {
--- a/pkg/blob/nodeserver.go
+++ b/pkg/blob/nodeserver.go
@@ -170,7 +170,7 @@ func (d *Driver) mountBlobfuseWithProxy(args, protocol string, authEnv []string)
 	mc := csiMetrics.NewCSIMetricContext("node_blobfuse_proxy_mount")
 	isOperationSucceeded := false
 	defer func() {
-		mc.ObserveWithLabels(isOperationSucceeded, Protocol, protocol)
+		mc.ObserveWithLabels(isOperationSucceeded, csiMetrics.Protocol, protocol)
 	}()
 
 	connectionTimeout := time.Duration(d.blobfuseProxyConnTimeout) * time.Second
@@ -213,7 +213,7 @@ func (d *Driver) mountBlobfuseInsideDriver(args string, protocol string, authEnv
 	mc := csiMetrics.NewCSIMetricContext("node_blobfuse_mount")
 	isOperationSucceeded := false
 	defer func() {
-		mc.ObserveWithLabels(isOperationSucceeded, Protocol, protocol)
+		mc.ObserveWithLabels(isOperationSucceeded, csiMetrics.Protocol, protocol)
 	}()
 
 	var cmd *exec.Cmd
@@ -354,11 +354,10 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 	mc := csiMetrics.NewCSIMetricContext("node_stage_volume").WithBasicVolumeInfo(d.cloud.ResourceGroup, "", d.Name)
 	isOperationSucceeded := false
 	defer func() {
-		mc.WithAdditionalVolumeInfo(VolumeID, volumeID)
-		mc.ObserveWithLabels(isOperationSucceeded,
-			"protocol", protocol,
-			"storage_account_type", blobStorageAccountType,
-			"is_hns_enabled", strconv.FormatBool(isHnsEnabled))
+		mc.WithAdditionalVolumeInfo(VolumeID, volumeID).ObserveWithLabels(isOperationSucceeded,
+			csiMetrics.Protocol, protocol,
+			csiMetrics.StorageAccountType, blobStorageAccountType,
+			csiMetrics.IsHnsEnabled, strconv.FormatBool(isHnsEnabled))
 	}()
 	mnt, err := d.ensureMountPoint(targetPath, fs.FileMode(mountPermissions))
 	if err != nil {
@@ -561,8 +560,7 @@ func (d *Driver) NodeUnstageVolume(_ context.Context, req *csi.NodeUnstageVolume
 	mc := csiMetrics.NewCSIMetricContext("node_unstage_volume").WithBasicVolumeInfo(d.cloud.ResourceGroup, "", d.Name)
 	isOperationSucceeded := false
 	defer func() {
-		mc.WithAdditionalVolumeInfo(VolumeID, volumeID)
-		mc.Observe(isOperationSucceeded)
+		mc.WithAdditionalVolumeInfo(VolumeID, volumeID).Observe(isOperationSucceeded)
 	}()
 
 	klog.V(2).Infof("NodeUnstageVolume: volume %s unmounting on %s", volumeID, stagingTargetPath)
@@ -621,8 +619,7 @@ func (d *Driver) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVolumeS
 	mc := csiMetrics.NewCSIMetricContext("node_get_volume_stats").WithBasicVolumeInfo(d.cloud.ResourceGroup, "", d.Name)
 	isOperationSucceeded := false
 	defer func() {
-		mc.WithAdditionalVolumeInfo(VolumeID, req.VolumeId)
-		mc.Observe(isOperationSucceeded)
+		mc.WithAdditionalVolumeInfo(VolumeID, req.VolumeId).Observe(isOperationSucceeded)
 	}()
 
 	if _, err := os.Lstat(req.VolumePath); err != nil {
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -27,6 +27,11 @@ import (
 
 const (
 	subSystem = "blob_csi_driver"
+
+	// Label keys for metrics
+	Protocol           = "protocol"
+	StorageAccountType = "storage_account_type"
+	IsHnsEnabled       = "is_hns_enabled"
 )
 
 var (
@@ -49,7 +54,7 @@ var (
 			Buckets:        []float64{0.1, 0.2, 0.5, 1, 5, 10, 15, 20, 30, 40, 50, 60, 100, 200, 300},
 			StabilityLevel: metrics.ALPHA,
 		},
-		[]string{"operation", "success", "protocol", "storage_account_type", "is_hns_enabled"},
+		[]string{"operation", "success", Protocol, StorageAccountType, IsHnsEnabled},
 	)
 
 	operationTotal = metrics.NewCounterVec(
@@ -144,9 +149,9 @@ func (mc *CSIMetricContext) Observe(success bool) {
 
 	// Record detailed metrics if labels are present
 	if len(mc.labels) > 0 {
-		protocol := mc.labels["protocol"]
-		storageAccountType := mc.labels["storage_account_type"]
-		isHnsEnabled := mc.labels["is_hns_enabled"]
+		protocol := mc.labels[Protocol]
+		storageAccountType := mc.labels[StorageAccountType]
+		isHnsEnabled := mc.labels[IsHnsEnabled]
 
 		operationDurationWithLabels.WithLabelValues(
 			mc.operation,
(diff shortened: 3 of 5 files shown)


pkg/blob/nodeserver.go
```diff
@@ -375,6 +377,11 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 		return &csi.NodeStageVolumeResponse{}, nil
 	}
 
+	mc := metrics.NewMetricContext(blobCSIDriverName, requestName, d.cloud.ResourceGroup, "", d.Name)
```
landreasyan: This change is duplicating the metrics. I don't think we need to add the cloud provider metrics back. Can you add any missing information to the csi metrics?

pkg/blob/controllerserver.go
```diff
@@ -564,10 +576,9 @@ func (d *Driver) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest)
 		}
 	}
 
-	csiMC := csiMetrics.NewCSIMetricContext("controller_delete_volume")
-	isOperationSucceeded := false
+	mc := metrics.NewMetricContext(blobCSIDriverName, requestName, d.cloud.ResourceGroup, d.cloud.SubscriptionID, d.Name)
```
landreasyan: Thinking more about this, the request name here should probably be `DeleteBlobContainer`. These are cloud provider metrics, so passing the actual operation name would make more sense.

landreasyan: Actually, these metrics are already included in the `DeleteBlobContainer` method with the correct operation name. This metrics is a duplicate.

pkg/blob/controllerserver.go
```diff
@@ -395,12 +403,9 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	defer d.volumeLocks.Release(volName)
 
 	var volumeID string
-	csiMC := csiMetrics.NewCSIMetricContext(requestName)
-	isOperationSucceeded := false
+	mc := metrics.NewMetricContext(blobCSIDriverName, requestName, d.cloud.ResourceGroup, d.cloud.SubscriptionID, d.Name)
```
landreasyan: This metrics already exists in `CreateBlobContainer`.

andyzhangx: we need a metric for the whole `controller_create_volume` e2e for csi driver including search account, create account and create blob container

pkg/metrics/metrics.go
```diff
@@ -118,18 +127,23 @@ func (mc *CSIMetricContext) Observe(success bool) {
...
+	keysAndValues = append(keysAndValues, additionalKeysAndValues...)
+	logger.V(int(mc.logLevel)).Info("Observed Request Latency", keysAndValues...)
 }
 
 // ObserveWithLabels records the operation with provided label pairs
-func (mc *CSIMetricContext) ObserveWithLabels(success bool, labelPairs ...string) {
-	if len(labelPairs)%2 != 0 {
-		// Invalid label pairs, just observe without labels
-		mc.Observe(success)
+func (mc *CSIMetricContext) ObserveWithLabels(success bool, labelPairs []string, additionalKeysAndValues ...interface{}) {
```
landreasyan: why not merge these with the existing labelPairs?

pkg/metrics/metrics.go
```diff
@@ -69,17 +72,21 @@ func init() {
...
+	operation  string
+	attributes []string
+	start      time.Time
+	labels     map[string]string
+	logLevel   int32
 }
 
 // NewCSIMetricContext creates a new CSI metric context
-func NewCSIMetricContext(operation string) *CSIMetricContext {
+func NewCSIMetricContext(operation, resourceGroup, subscriptionID, source string) *CSIMetricContext {
```
landreasyan: Would recommend passing the disk URI only as it includes the rest of the information and to make it optional.

landreasyan: Actually, it might make more sense to pass the disk uri in to the observe method, not here.

pkg/metrics/metrics.go
```diff
@@ -93,11 +100,13 @@ func (mc *CSIMetricContext) WithLabel(key, value string) *CSIMetricContext {
...
+func (mc *CSIMetricContext) Observe(success bool, additionalKeysAndValues ...interface{}) {
 	duration := time.Since(mc.start).Seconds()
-	successStr := "false"
-	if success {
-		successStr = "true"
+	successStr := "true"
+	resultCode := "succeeded"
+	if !success {
+		successStr = "false"
+		resultCode = "failed_" + mc.operation
```
landreasyan: not sure we need to append the operation name here. The operation name is logged already, the result code should indicate either success or failure.

pkg/metrics/metrics.go
```diff
@@ -118,18 +127,23 @@ func (mc *CSIMetricContext) Observe(success bool) {
 			isHnsEnabled,
 		).Observe(duration)
 	}
+
```
landreasyan, line 164: Would recommend refactoring into a separate method of `CSIMetricContext`, `logLatency`. I don't think it needs to replicate the cloud provider log 1-1. Would recommend adding the log line similar to `CSI Operation Completed operation=node_publish_volume, latency_seconds=0.234, volume=<uri>, success=true`.

pkg/metrics/metrics.go
```diff
@@ -69,17 +72,21 @@ func init() {
...
+func NewCSIMetricContext(operation, resourceGroup, subscriptionID, source string) *CSIMetricContext {
 	return &CSIMetricContext{
-		operation: operation,
-		start:     time.Now(),
-		labels:    make(map[string]string),
+		operation:  operation,
+		attributes: []string{strings.ToLower(resourceGroup), subscriptionID, source},
+		start:      time.Now(),
+		labels:     make(map[string]string),
+		logLevel:   3,
```
landreasyan: You can add a `WithLogLevel` method to `CSIMetricContext` to allow configuring the level per csi call, where the 3 here would be the default value.

pkg/metrics/metrics.go
```diff
@@ -17,10 +17,13 @@ limitations under the License.
 package metrics
 
 import (
+	"strings"
 	"time"
 
 	"k8s.io/component-base/metrics"
 	"k8s.io/component-base/metrics/legacyregistry"
+
+	"sigs.k8s.io/cloud-provider-azure/pkg/log"
```
landreasyan: Don't add dependency to cloud provider package just for the logs. Use `"k8s.io/klog/v2"`

pkg/metrics/metrics.go
landreasyan: If in case of blob the volume if doesn't already include all this info, would recommend adding the following method to pass these parameters. 

```
// WithVolumeInfo sets volume-specific information

func (mc *CSIMetricContext) WithVolumeInfo(volumeID, subscriptionID, resourceGroup, blobName string) *CSIMetricContext {
    mc.volumeContext["volume_id"] = volumeID
    mc.volumeContext["subscription_id"] = subscriptionID
    mc.volumeContext["resource_group"] = resourceGroup
    mc.volumeContext["blob_name"] = blobName
    return mc
}```

File: pkg/blob/nodeserver.go [hunk 10]
```diff
@@ -351,13 +351,13 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 		return nil, status.Errorf(codes.InvalidArgument, "fsGroupChangePolicy(%s) is not supported, supported fsGroupChangePolicy list: %v", fsGroupChangePolicy, supportedFSGroupChangePolicyList)
 	}
 
-	mc := csiMetrics.NewCSIMetricContext("node_stage_volume")
+	mc := csiMetrics.NewCSIMetricContext("node_stage_volume", d.cloud.ResourceGroup, "", d.Name)
```
landreasyan: Instead of passing these parameters directly, use a member method as follows: 

```
mc := csiMetrics.NewCSIMetricContext("node_get_volume_stats").
    WithVolumeInfo(volumeID, subscriptionID, resourceGroup, blobName)
```
These values will not make sense for every csi method, so it makes sense to add them only when needed.

pkg/blob/nodeserver.go
```diff
@@ -616,10 +616,10 @@ func (d *Driver) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVolumeS
 		return resp, nil
 	}
 
-	mc := csiMetrics.NewCSIMetricContext("node_get_volume_stats")
+	mc := csiMetrics.NewCSIMetricContext("node_get_volume_stats", d.cloud.ResourceGroup, "", d.Name)
```
landreasyan: Add and use `WithLogLevel` method to increase the log level for methods that we don't normally need the latency info in prod scenarios.

pkg/metrics/metrics.go
landreasyan: or ```"operation=node_publish_volume, latency_seconds=0.234, volume_id=...,         subscription_id=..., resource_group=..., blob_name=..., success=true"```

pkg/metrics/metrics.go
hccheng72: because additionalKeysAndValues are just for the logging purpose, may not be appropriate for a metrics label (e.g. VolumeId).

pkg/metrics/metrics.go
hccheng72: This is the format from original log. Your suggestion makes sense to me but not sure if there're any querying purposes. @andyzhangx, could you please confirm this?

hccheng72: @andyzhangx, a couple of things in the log to confirm whether it's acceptable for you, using examples below:

1. changing from `result_code="succeeded"` to `success="true"`

1. changing from `result_code="failed_csi_driver_node_expand_volume"` to `success="false"` (won't repeat operation since it is already in the log)

pkg/blob/controllerserver.go
```diff
@@ -395,12 +395,11 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	defer d.volumeLocks.Release(volName)
 
 	var volumeID string
-	csiMC := csiMetrics.NewCSIMetricContext(requestName)
+	csiMC := csiMetrics.NewCSIMetricContext(requestName).WithBasicVolumeInfo(d.cloud.ResourceGroup, d.cloud.SubscriptionID, d.Name)
```
landreasyan, line 398: Can we add the volume ID here as well? Does volume id include any info not already covered by these parameters?

pkg/metrics/metrics.go
```diff
@@ -69,20 +71,46 @@ func init() {
...
 	return &CSIMetricContext{
-		operation: operation,
-		start:     time.Now(),
-		labels:    make(map[string]string),
+		operation:     operation,
+		volumeContext: make(map[string]string),
+		start:         time.Now(),
+		labels:        make(map[string]string),
+		logLevel:      3,
+		enableLogging: true,
```
landreasyan: You can achieve this by changing the log levels. We probably don't need another parameter to enable and disable logging.

pkg/blob/controllerserver.go
hccheng72: We don’t pass the volume ID every time (including the cases in other azure csi drivers), and in some cases it may be a different key-value pair. For additional information like this, we would need to pass both the key and the value explicitly. If we combine this with the basic information, it means we’d have to pass keys such as "resource_group", "subscription_id", and "source" every time. That feels a bit messy.

pkg/metrics/metrics.go
andyzhangx: I would suggest using one unified logging format for all csi drivers, e.g. for disk driver, we are already using following format:
```
I0209 15:49:25.360706       1 azure_metrics.go:106] "Observed Request Latency" logger="logLatency" latency_seconds=7.028497647 request="azuredisk_csi_driver_controller_unpublish_volume" resource_group="capz-1lr3id" subscription_id="46678f10-4bbb-447e-98e8-d2829589f2d8" source="disk.csi.azure.com" volumeid="/subscriptions/46678f10-4bbb-447e-98e8-d2829589f2d8/resourceGroups/capz-1lr3id/providers/Microsoft.Compute/disks/reattach-disk-multiple-nodes" node="capz-1lr3id-mp-0000001" result_code="succeeded"
```

hccheng72: That makes sense. Once we finalize the metrics package here, we can apply it to the other CSI drivers, ensuring that all existing logs are preserved and consistency is maintained across drivers. 
cc: @landreasyan

pkg/blob/nodeserver.go
```diff
@@ -557,9 +558,10 @@ func (d *Driver) NodeUnstageVolume(_ context.Context, req *csi.NodeUnstageVolume
 	}
 	defer d.volumeLocks.Release(lockKey)
 
-	mc := csiMetrics.NewCSIMetricContext("node_unstage_volume")
+	mc := csiMetrics.NewCSIMetricContext("node_unstage_volume").WithBasicVolumeInfo(d.cloud.ResourceGroup, "", d.Name)
```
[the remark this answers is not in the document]
hccheng72: This is from original code.

pkg/blob/nodeserver.go
```diff
@@ -616,9 +618,10 @@ func (d *Driver) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVolumeS
 		return resp, nil
 	}
 
-	mc := csiMetrics.NewCSIMetricContext("node_get_volume_stats")
+	mc := csiMetrics.NewCSIMetricContext("node_get_volume_stats").WithBasicVolumeInfo(d.cloud.ResourceGroup, "", d.Name)
```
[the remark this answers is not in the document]
hccheng72: same as original code

pkg/metrics/metrics.go
```diff
@@ -69,20 +71,44 @@ func init() {
...
 	return &CSIMetricContext{
-		operation: operation,
-		start:     time.Now(),
-		labels:    make(map[string]string),
+		operation:     operation,
+		volumeContext: make(map[string]string),
+		start:         time.Now(),
+		labels:        make(map[string]string),
+		logLevel:      3,
 	}
```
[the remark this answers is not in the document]
hccheng72: same as original setting

andyzhangx (commented):
/retest

pkg/metrics/metrics.go
```diff
@@ -69,20 +71,44 @@ func init() {
...
 }
 
+// WithBasicVolumeInfo adds the standard volume-related context to the metric context
+func (mc *CSIMetricContext) WithBasicVolumeInfo(resourceGroup, subscriptionID, source string) *CSIMetricContext {
+	mc.volumeContext["resource_group"] = strings.ToLower(resourceGroup)
+	mc.volumeContext["subscription_id"] = subscriptionID
```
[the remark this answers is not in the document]
hccheng72: same as original code

PR #2430, opened 2026-04-27
[WIP] feat: default to kubelet identity when AzureStorageIdentityClientID is empty for MSI auth

andyzhangx:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it:**
When `AzureStorageAuthType` is set to `MSI` but `AzureStorageIdentityClientID` is not specified, the driver now:

1. **Defaults to the kubelet identity** (`UserAssignedIdentityID` from cloud config) — this is the built-in managed identity bound to the AKS agent node pool (named `{cluster-name}-agentpool`)
2. **Returns a clear error** if the kubelet identity is also not available, instead of silently proceeding and failing at mount time with a cryptic error

This improves the user experience for the common case where users set `AzureStorageAuthType: MSI` in their PV/StorageClass but forget to specify the identity client ID — the driver will automatically use the kubelet identity if available.

**How does this PR make you feel?**
Less cryptic mount failures 🎉

**Does this PR introduce a user-facing change?**
```release-note
When AzureStorageAuthType is set to MSI and AzureStorageIdentityClientID is not specified, the driver defaults to the kubelet identity. If no identity is available, a clear error is returned.
```

### DIFF
--- a/pkg/blob/blob.go
+++ b/pkg/blob/blob.go
@@ -754,14 +754,18 @@ func (d *Driver) GetAuthEnv(ctx context.Context, volumeID, protocol string, attr
 		authEnv = append(authEnv, "AZURE_STORAGE_SPN_TENANT_ID="+storageSPNTenantID)
 	}
 
-	if azureStorageAuthType == storageAuthTypeMSI {
-		// check whether authEnv contains AZURE_STORAGE_IDENTITY_ prefix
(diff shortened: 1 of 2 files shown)


mittachaitu:
Hi @andyzhangx what will be the impact for inline volumes? Does any arbitrary pod without right permissions can use kubelet identity to mount the container?

nearora-msft:
@andyzhangx Is there more work pending on this PR?

PR #2460, opened 2026-05-14
fix: harden ephemeral mount option handling for inline volumes.

kapilupadhayay:
Harden ephemeral mount option handling for inline volumes.

**What type of PR is this?**
/kind bug

Blocks the users to specify --tmp-path on ephemeral volumes

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/blob/blob.go
+++ b/pkg/blob/blob.go
@@ -1341,6 +1341,13 @@ func SanitizeMountOptions(mountOptions []string) ([]string, error) {
 			if strings.IndexFunc(rest, unicode.IsSpace) >= 0 {
 				return nil, fmt.Errorf("mount option %q: FUSE -o options must not contain whitespace", trimmed)
 			}
+			// FUSE -o options (allow_other, ro, rw, umask=…, uid=…, …) never start
+			// with "-". A leading "-" means the user is smuggling a blobfuse2 CLI
+			// flag through the passthrough, which the proxy will hand to blobfuse2
+			// as a top-level argument after splitting on whitespace.
+			if strings.HasPrefix(rest, "-") {
+				return nil, fmt.Errorf("mount option %q: -o values must be FUSE options, not CLI flags", trimmed)
+			}
 			sanitized = append(sanitized, trimmed)
 			continue
 		}
@@ -1353,6 +1360,14 @@ func SanitizeMountOptions(mountOptions []string) ([]string, error) {
 		// Validate enum-typed flags when a value is present.
 		if len(parts) == 2 {
 			flagValue := parts[1]
+			// Reject any whitespace in the value. The proxy joins all tokens
+			// with single spaces and then splits on " " to build argv; any
+			// whitespace surviving inside a value (e.g. "--cache-size-mb=512,-o --config-file=/tmp/x")
+			// is normalised to a space by TrimDuplicatedSpace and split off
+			// into an injected argv element (here: "--config-file=/tmp/x").
+			if strings.IndexFunc(flagValue, unicode.IsSpace) >= 0 {
+				return nil, fmt.Errorf("mount option %q: value must not contain whitespace", trimmed)
+			}
 			switch flagName {
 			case "--log-level":
 				if _, ok := allowedLogLevels[flagValue]; !ok {
@@ -1408,13 +1423,19 @@ func appendDefaultMountOptions(mountOptions []string, tmpPath, containerName str
 	allMountOptions := mountOptions
 
 	for k, v := range defaultMountOptions {
-		if _, isIncluded := included[k]; !isIncluded {
-			if v != "" {
-				allMountOptions = append(allMountOptions, fmt.Sprintf("%s=%s", k, v))
-			} else {
-				allMountOptions = append(allMountOptions, k)
-			}
+		if included[k] {
+			continue
 		}
+		if v == "" {
+			// Defensive: a bare default flag would consume the next argv element
+			// as its value when blobfuse2 parses the split args string. None of
+			// the current defaults legitimately use empty values; if a future
+			// default needs a bare flag, move it to a separate boolean-flags
+			// slice rather than going through this map.
+			klog.Warningf("skipping default mount option %q with empty value", k)
+			continue
+		}
+		allMountOptions = append(allMountOptions, fmt.Sprintf("%s=%s", k, v))
 	}
 
 	return allMountOptions
--- a/pkg/blob/blob.go
+++ b/pkg/blob/blob.go
@@ -1411,10 +1411,9 @@ func appendDefaultMountOptions(mountOptions []string, tmpPath, containerName str
 		if _, isIncluded := included[k]; !isIncluded {
 			if v != "" {
 				allMountOptions = append(allMountOptions, fmt.Sprintf("%s=%s", k, v))
+			} else {
+				allMountOptions = append(allMountOptions, k)
 			}
-			// v == "": skip entirely — do NOT emit a bare flag (e.g. "--container-name"
-			// with no value). A bare GNU-style flag causes blobfuse2 to consume the next
-			// argv element as its value, silently discarding a driver-controlled flag.
 		}
 	}
 
--- a/pkg/blob/blob.go
+++ b/pkg/blob/blob.go
@@ -1411,9 +1411,10 @@ func appendDefaultMountOptions(mountOptions []string, tmpPath, containerName str
 		if _, isIncluded := included[k]; !isIncluded {
 			if v != "" {
 				allMountOptions = append(allMountOptions, fmt.Sprintf("%s=%s", k, v))
-			} else {
-				allMountOptions = append(allMountOptions, k)
 			}
+			// v == "": skip entirely — do NOT emit a bare flag (e.g. "--container-name"
+			// with no value). A bare GNU-style flag causes blobfuse2 to consume the next
+			// argv element as its value, silently discarding a driver-controlled flag.
 		}
 	}
 
--- a/pkg/blob/blob.go
+++ b/pkg/blob/blob.go
@@ -178,6 +178,84 @@ var (
 	defaultAzureOAuthTokenDir = "/var/lib/kubelet/plugins/" + DefaultDriverName
 
 	subscriptionIDRegex = regexp.MustCompile(`^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$`)
+
+	// containerNameRegex enforces Azure Blob Storage container naming rules.
+	// Names containing spaces or special characters cannot match and are
+	// rejected before reaching the blobfuse2 args string.
+	containerNameRegex = regexp.MustCompile(`^[a-z0-9][a-z0-9-]{1,61}[a-z0-9]$`)
+
+	// allowedEphemeralMountOptions is the allowlist of blobfuse2 CLI flags that
+	// may appear in volumeAttributes.mountOptions for an ephemeral inline volume.
+	// Any flag NOT in this map is rejected by SanitizeMountOptions.
+	//
+	// Verified against: blobfuse2 mount --help (v2.x)
+	//
+	// Intentionally absent (must remain driver-controlled):
+	//   --tmp-path          — redirects file-cache root to arbitrary host path
+	//   --block-cache-path  — same primitive for block-cache mode
+	//   --config-file       — loads arbitrary blobfuse2 config from host
+	//   --log-file-path     — root-owned log writes to arbitrary host path
+	//   --container-name    — driver sets this from the validated containerName
+	//                         volumeAttribute; allowing it in mountOptions would
+	//                         override that validation entirely
+	//   --passphrase        — decrypts --secure-config files; enables config-file
+	//                         attack path indirectly
+	//   --secure-config     — enables encrypted config loading; combined with
+	//                         --passphrase reinstates the --config-file attack
+	//
+	// Note: FUSE passthrough tokens "-o <option>[=value]" are always permitted;
+	// they are handled separately in SanitizeMountOptions.
+	allowedEphemeralMountOptions = map[string]struct{}{
+		"--allow-other":                    {},
+		"--attr-cache-timeout":             {},
+		"--attr-timeout":                   {},
+		"--block-cache":                    {},
+		"--block-cache-block-size":         {},
+		"--block-cache-disk-size":          {},
+		"--block-cache-disk-timeout":       {},
+		"--block-cache-parallelism":        {},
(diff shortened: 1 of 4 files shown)


pkg/blob/blob.go
```diff
@@ -1110,6 +1110,43 @@ func (d *Driver) useDataPlaneAPI(ctx context.Context, volumeID, accountName stri
 	return false
 }
 
+// blockedEphemeralMountOptions is the set of blobfuse2 options that users must not be able to
+// override via volumeAttributes.mountOptions on inline ephemeral volumes.
+//
+// --tmp-path: primary exploit vector. blobfuse2 writes blob content (attacker-controlled) into
```
andyzhangx: `--tmp-path` mount option is needed in blobfuse2 mount, what about limit the value, e.g. only `/tmp` or `/mnt` can be used in `--tmp-path` mount option

andyzhangx (commented):
can you sign the easycla first? (select individual contributor option)

pkg/blob/blob.go
andyzhangx: nvm, the default `/mnt` should be ok, resolve this comment now

pkg/blob/blob.go
```diff
@@ -1110,6 +1110,43 @@ func (d *Driver) useDataPlaneAPI(ctx context.Context, volumeID, accountName stri
...
+// this directory as root. Setting it to e.g. /etc/kubernetes/manifests allows the attacker to
+// land an arbitrary static Pod manifest, achieving host-level RCE.
+//
+// --config-file: two-step bypass for the --tmp-path block. An attacker can upload a blobfuse2
+// config YAML as a blob (step 1, normal mount caches it to /mnt/<volumeID>/evil.yaml), then
+// point a second ephemeral mount at that cached file via --config-file, which sets tmp-path
+// inside the config, reinstating the RCE path. The driver always generates its own safe config.
+var blockedEphemeralMountOptions = []string{
+	"--tmp-path",
```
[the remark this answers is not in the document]
kapilupadhayay: Fixed

pkg/blob/nodeserver.go
```diff
@@ -450,7 +450,10 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 	// Get mountOptions that the volume will be formatted and mounted with
 	mountOptions := mountFlags
 	if ephemeralVol {
-		mountOptions = util.JoinMountOptions(mountOptions, strings.Split(ephemeralVolMountOptions, ","))
+		// Sanitize user-supplied mount options before use: strip options that are
+		// security-sensitive (e.g. --tmp-path) and must be driver-controlled.
+		sanitized := sanitizeMountOptions(strings.Split(ephemeralVolMountOptions, ","))
```
landreasyan: Should it silently sanitize and mount or should it throw an error saying that it can't mount with provided options?

landreasyan: I think I would prefer this failing with `InvaidArgument` error as the user isn't getting the requested behavior.

pkg/blob/blob.go
```diff
@@ -1110,6 +1110,50 @@ func (d *Driver) useDataPlaneAPI(ctx context.Context, volumeID, accountName stri
...
+// sanitizeMountOptions removes options from the provided list that are in the
+// blockedEphemeralMountOptions denylist. It is called when processing user-supplied mount options
+// for inline ephemeral CSI volumes before the driver appends its own safe defaults.
+func sanitizeMountOptions(mountOptions []string) []string {
+	filtered := make([]string, 0, len(mountOptions))
+	for _, opt := range mountOptions {
+		blocked := false
+		for _, blockedPrefix := range blockedEphemeralMountOptions {
+			if strings.HasPrefix(strings.TrimSpace(opt), blockedPrefix) {
+				klog.Warningf("mount option %q is not allowed for ephemeral volumes and will be ignored", opt)
```
landreasyan: is there a `--temp-path*" mount that is ok? Any mount option starting with `temp-path` will be blocked in this code path.

landreasyan: what about the short form of these mount options?

pkg/blob/blob.go
```diff
@@ -1110,6 +1110,50 @@ func (d *Driver) useDataPlaneAPI(ctx context.Context, volumeID, accountName stri
...
+// persists downloaded blocks (attacker-controlled blob content) to this directory as root.
+// An ephemeral volume with --block-cache-path=/etc/kubernetes/manifests (optionally combined
+// with --block-cache=true) reproduces the same static-Pod-injection / host RCE primitive.
+//
+// --config-file: two-step bypass for the above blocks. An attacker can upload a blobfuse2
+// config YAML as a blob (step 1, normal mount caches it to /mnt/<volumeID>/evil.yaml), then
+// point a second ephemeral mount at that cached file via --config-file, which sets
+// file_cache.path or block_cache.path inside the config, reinstating the RCE path.
+var blockedEphemeralMountOptions = []string{
+	"--tmp-path",
```
landreasyan: should `--file-path` also be included?

landreasyan:
Instead of keeping a deny-list of mounts, should we have a list of allowed mount options?

NedAnd1:
> Instead of keeping a deny-list of mounts, should we have a list of allowed mount options?

Agreed, an allow list of safe mount options would improve security,
esp. since we're passing these options to a component that can add new ones at any time.

pkg/blob/nodeserver.go
kapilupadhayay: Yes, its better to fail with InvalidArg.

pkg/blob/blob.go
kapilupadhayay: Made the check exact. None of the options have shorter forms.

pkg/blob/blob.go
kapilupadhayay: could not find any --file-path arguments with mount or mountv1

kapilupadhayay:
> Instead of keeping a deny-list of mounts, should we have a list of allowed mount options?

I believe deny list is still better. It won't be a breaking change and new options when added in the future need not be maintained in the list.

andyzhangx:
@kapilupadhayay pls also add `log-file-path` parameter into the deny list, thx

pkg/blob/blob.go
```diff
@@ -1110,6 +1188,123 @@ func (d *Driver) useDataPlaneAPI(ctx context.Context, volumeID, accountName stri
...
+// a space. A space inside a value would be misinterpreted as a flag separator,
+// allowing arbitrary blobfuse2 flags to be passed. This check is a
+// last-resort safety net applied after all other validations
+// (SanitizeMountOptions, ValidateContainerName) and after
+// appendDefaultMountOptions has added driver-controlled defaults.
+func ValidateMountArgValues(mountOptions []string) error {
+	for _, opt := range mountOptions {
+		// Split on the first "=" to isolate the value.
+		parts := strings.SplitN(strings.TrimSpace(opt), "=", 2)
+		if len(parts) == 2 && strings.Contains(parts[1], " ") {
```
landreasyan: We might need to check strings.ContainsAny(parts[1], " \t\n\v\f\r")

pkg/blob/blob.go
```diff
@@ -178,6 +178,84 @@ var (
...
+	//                         volumeAttribute; allowing it in mountOptions would
+	//                         override that validation entirely
+	//   --passphrase        — decrypts --secure-config files; enables config-file
+	//                         attack path indirectly
+	//   --secure-config     — enables encrypted config loading; combined with
+	//                         --passphrase reinstates the --config-file attack
+	//
+	// Note: FUSE passthrough tokens "-o <option>[=value]" are always permitted;
+	// they are handled separately in SanitizeMountOptions.
+	allowedEphemeralMountOptions = map[string]struct{}{
```
landreasyan, line 209: driver seems to be adding `--use-adls` to mount options. Should it be added to the list?

pkg/blob/blob.go
```diff
@@ -178,6 +178,84 @@ var (
...
+		"--log-level":                      {},
+		"--log-type":                       {},
+		"--low-disk-threshold":             {},
+		"--negative-timeout":               {},
+		"--no-symlinks":                    {},
+		"--pool-size":                      {},
+		"--preload":                        {},
+		"--preserve-acl":                   {},
+		"--read-only":                      {},
+		"--subdirectory":                   {},
```
landreasyan, line 253: I think any flag that takes a string as an argument should be validated with a similar regex as the containerName or we should remove it from the allowlist.

pkg/blob/blob.go
kapilupadhayay: The driver inserts --use-adls=true itself at nodeserver.go, whenever isHnsEnabled == true. Users of ephemeral inline volumes don't set it — the driver detects it from the storage account. Putting it in the allowlist would let users force it on non-HNS accounts, which could cause mount failures or confusing behaviour.

pkg/blob/blob.go
kapilupadhayay: The remaining flags accepting strings were, (they were already safe as no host path access is associated with them). Still its good to add resolution for them:

1. --subdirectory — Azure blob prefix path (e.g. images/2024/january) -> Regex for blobpath is added.
2. --filter — blob name filter pattern (e.g. *.txt, images/*.jpg) -> Regex Added
3.  --log-type -> Set of valid values added .
4. --log-levels -> Set of valid values added.

pkg/blob/blob.go
```diff
@@ -1110,6 +1188,123 @@ func (d *Driver) useDataPlaneAPI(ctx context.Context, volumeID, accountName stri
...
+			return fmt.Errorf("mount option %q has an invalid value: spaces are not permitted", parts[0])
+		}
+	}
+	return nil
+}
+
+// ValidateContainerName rejects any containerName that does not match the
+// Azure Blob Storage naming rules. This also prevents invalid names from
+// value containing spaces cannot match the regex and is rejected before it
+// reaches the blobfuse2 args string that the proxy splits on whitespace.
```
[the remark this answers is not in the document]
kapilupadhayay: Done

pkg/blob/blob.go
```diff
@@ -1110,6 +1188,123 @@ func (d *Driver) useDataPlaneAPI(ctx context.Context, volumeID, accountName stri
...
+func ValidateContainerName(containerName string) error {
+	// Empty is allowed: some flows (e.g. workload identity) resolve the container
+	// name later; an empty value cannot inject any arguments.
+	if containerName == "" {
+		return nil
```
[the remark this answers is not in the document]
kapilupadhayay: Fixed

pkg/blob/blob.go
```diff
@@ -1110,6 +1218,169 @@ func (d *Driver) useDataPlaneAPI(ctx context.Context, volumeID, accountName stri
...
+		for _, t := range strings.Split(s, ",") {
+			if t = strings.TrimSpace(t); t != "" {
+				if strings.HasPrefix(t, "-o ") {
+					// The space between "-o" and its value is expected; reject
+					// any further whitespace embedded in the value part.
+					if strings.IndexFunc(t[3:], unicode.IsSpace) >= 0 {
+						return nil
+					}
+				} else if strings.IndexFunc(t, unicode.IsSpace) >= 0 {
+					return nil
```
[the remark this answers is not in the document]
kapilupadhayay: Fixed

pkg/blob/blob.go
```diff
@@ -1110,6 +1218,169 @@ func (d *Driver) useDataPlaneAPI(ctx context.Context, volumeID, accountName stri
...
+// whitespace. Whitespace inside a value would be misinterpreted as a flag
+// separator by the blobfuse-proxy or blobfuse2's own argument parser. This
+// check is a last-resort safety net applied after all other validations
+// (SanitizeMountOptions, ValidateContainerName) and after
+// appendDefaultMountOptions has added driver-controlled defaults.
+func ValidateMountArgValues(mountOptions []string) error {
+	for _, opt := range mountOptions {
+		// Split on the first "=" to isolate the value.
+		parts := strings.SplitN(strings.TrimSpace(opt), "=", 2)
+		if len(parts) == 2 && strings.ContainsAny(parts[1], " \t\n\v\f\r") {
```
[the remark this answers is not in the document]
kapilupadhayay: Fixed

kapilupadhayay:
@landreasyan @andyzhangx would you please approve the request.

nearora-msft (approved):

sunpa93 (approved):
/lgtm

pkg/blob/blob.go
```diff
@@ -1140,9 +1411,10 @@ func appendDefaultMountOptions(mountOptions []string, tmpPath, containerName str
 		if _, isIncluded := included[k]; !isIncluded {
 			if v != "" {
 				allMountOptions = append(allMountOptions, fmt.Sprintf("%s=%s", k, v))
-			} else {
-				allMountOptions = append(allMountOptions, k)
 			}
+			// v == "": skip entirely — do NOT emit a bare flag (e.g. "--container-name"
```
andyzhangx: there are mount option like `-o allow_other` which does not have a k/v pair, we should allow such mount option, and this change is irrelevant to the original PR, pls revert this.

kapilupadhayay: This change is reverted now. I have added a more  scoped check for containerNameField being "" in ephemeral case to take care of the earlier copilot review recommendation.

pkg/blob/nodeserver.go
```diff
@@ -485,6 +481,13 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 	}
 	mountOptions = appendDefaultMountOptions(mountOptions, tmpPath, containerName)
 
+	// Last-resort check: ensure no value in the final mount options contains a
+	// space. The proxy splits the flat args string on spaces; a space inside any
+	// value would be treated as a flag separator.
+	if err := ValidateMountArgValues(mountOptions); err != nil {
```
andyzhangx: this would still possibly be broken on existing cases, would you also revert this since it does not fix the inline volume issue.

pkg/blob/blob.go
andyzhangx: also revert these comments

andyzhangx (commented):
note: I will cherrypick this PR to release-1.27 and 1.26, let's make sure this PR only fixes the inline volume issue, thanks!

pkg/blob/nodeserver.go
kapilupadhayay: Done

kapilupadhayay:
@andyzhangx , @landreasyan  all the comments addressed now. Please approve the PR.

andyzhangx (approved):
/lgtm

PR #2465, opened 2026-05-16
fix: stop ticker/timer in waitForMount to avoid leak

dimbleby:
**What type of PR is this?**
/kind bug

waitForMount used time.Tick and time.After. time.Tick's underlying ticker is never garbage collected per the Go documentation, so every blobfuse mount leaked a runtime ticker that kept firing after the function returned. On the timeout path the ticker would continue to tick forever.

Use time.NewTicker and time.NewTimer with deferred Stop so resources are released when the function returns.

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


andyzhangx (approved):
/lgtm

PR #2491, opened 2026-06-05
fix: scope container name validation to inline volume only

sunpa93:
**What type of PR is this?**
/kind bug

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```


andyzhangx:
@sunpa93 this fix is incomplete, it still breaks in inline volume scenario, here is the complete fix: https://github.com/kubernetes-sigs/blob-csi-driver/pull/2502

sunpa93:
@andyzhangx That was our intention for the hotfix: take a conservative approach and to scope the validation to inline volume only. The plan is to further validate all scenarios where $ needs to be used in container name first and discuss a proper fix.

PR #2502, opened 2026-06-10
fix: allow Azure system containers ($web, $root, $logs) in ValidateContainerName

andyzhangx:
## What type of PR is this?
/kind bug

## What this PR does / why we need it

Fixes `ValidateContainerName` to allow Azure Blob Storage system containers that use a `$` prefix (e.g. `$web`, `$root`, `$logs`, `$blobchangefeed`).

Instead of maintaining a hardcoded whitelist of known system container names, the validation now allows **any** container name starting with `$`. This is safe because:
1. Users cannot create containers with `$` prefix (Azure API rejects them)
2. The `$` prefix is Azure's reserved marker for system containers
3. Azure may add new system containers in the future without requiring driver updates

### Changes
- Replace the `azureSystemContainers` map with a simple `strings.HasPrefix(containerName, "$")` check
- Add test case for `$blobchangefeed` to verify future-proofing

### References
- [Azure container naming rules](https://learn.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata#container-names)

Signed-off-by: Andy Zhang <andyzhangx@live.com>


NedAnd1:
@andyzhangx there are more special container names than the ones you called out & they could change over time, but they all start with a dollar sign, so maybe we could change `containerNameRegex` to allow a dollar sign in the beginning?
e.g. `^\$?[a-z0-9][a-z0-9-]+[a-z0-9]$`

PR #2506, opened 2026-06-11
fix: allow --use-adls mount option for ephemeral volumes

### LINKED ISSUE #2505: --use-adls no long works on ephemeral volumes as of v1.27.6
**What happened**: After upgrading our AKS cluster from 1.33 to 1.34 our workloads no longer mount blob storage accounts. They are getting the error: 

```
MountVolume.SetUp failed for volume "dev-blob-storage" : rpc error: code = InvalidArgument desc = mount option "--use-adls" is not allowed for ephemeral volumes
```

**What you expected to happen**: This function should continue to work

**How to reproduce it**: 

```
apiVersion: v1
kind: Pod
metadata:
  name: blob-csi-pod
spec:
  containers:
    - name: app
      image: busybox
      command: ["sleep", "infinity"]
      volumeMounts:
        - name: blob-volume
          mountPath: /mnt/blob
  volumes:
    - name: blob-volume
      csi:
        driver: blob.csi.azure.com
        volumeAttributes:
          protocol: fuse2
          containerName: my-container
          secretName: my-secret
          mountOptions: "-o allow_other,--file-cache-timeout-in-seconds=120,--cancel-list-on-mount-seconds=10,-o attr_timeout=120,-o entry_timeout=120,-o negative_timeout=120,--log-level=LOG_WARNING,--use-adls=true"
```

**Anything else we need to know?**:  This worked in 1.26.11.
**Environment**:

- CSI Driver version: v1.27.6
- Kubernetes version (use `kubectl version`): 1.34.8
- OS (e.g. from /etc/os-release): Azure Linux
- Kernel (e.g. `uname -a`): Unknown
- Install tools:
- Others:

jabbera:

PR #2476 introduced an allowlist of blobfuse2 flags permitted in `volumeAttributes.mountOptions` for ephemeral inline volumes (`allowedEphemeralMountOptions`). The `--use-adls` flag was omitted from that allowlist, so workloads mounting ADLS Gen2 / HNS accounts via inline volumes started failing with:

```
mount option "--use-adls" is not allowed for ephemeral volumes
```

`--use-adls` only toggles ADLS Gen2 (hierarchical namespace) semantics; it does not redirect host paths (`--tmp-path`, `--block-cache-path`), load arbitrary config (`--config-file`), or override driver-controlled values (`--container-name`). It therefore meets the same safety criteria as the other allowlisted flags. Notably, the driver itself appends `--use-adls=true` when HNS is detected, confirming it is a trusted flag.

This adds `--use-adls` to `allowedEphemeralMountOptions` and covers it with a unit test.

Fixes #2505

Regression introduced in #2476.

**Release note**:
```release-note
Fix `--use-adls` mount option being rejected on ephemeral inline volumes.
```

### DIFF
--- a/pkg/blob/blob.go
+++ b/pkg/blob/blob.go
@@ -252,6 +252,7 @@ var (
 		"--read-only":                      {},
 		"--subdirectory":                   {},
 		"--sync-to-flush":                  {},
+		"--use-adls":                       {},
 		"--use-attr-cache":                 {},
 		"--virtual-directory":              {},
 		"--wait-for-mount":                 {},
--- a/pkg/blob/blob_test.go
+++ b/pkg/blob/blob_test.go
@@ -1657,6 +1657,12 @@ func TestSanitizeMountOptions(t *testing.T) {
 			wantErr:  false,
 			expected: []string{"--cancel-list-on-mount-seconds=60"},
 		},
+		{
+			name:     "--use-adls is allowed (used for ADLS Gen2/HNS accounts, see issue #2505)",
+			options:  []string{"--use-adls=true"},
+			wantErr:  false,
+			expected: []string{"--use-adls=true"},
+		},
 		// -o FUSE passthrough: a space after -o would be treated as additional flags
 		// by the proxy, so the value must be a single token.
 		{


andyzhangx (commented):
/ok-to-test

andyzhangx (approved):
/lgtm

jabbera:
@andyzhangx thank you. Query: when will I get this in my cluster? I don't install this driver manually. It's managed by the AKS infrastructure.

andyzhangx:
> @andyzhangx thank you. Query: when will I get this in my cluster? I don't install this driver manually. It's managed by the AKS infrastructure.

that would be in next AKS release, pls email me your subs, cluster version, we could revert to the old working version early next week.

PR #2558, opened 2026-07-16
feat: allow blobfuse2 attr-cache-max-size-mb and kernel-list-cache-ti…

kapilupadhayay:
…meout on ephemeral volumes

Add --attr-cache-max-size-mb and --kernel-list-cache-timeout (both introduced in blobfuse2 v2.5.4) to allowedEphemeralMountOptions so they pass through SanitizeMountOptions for ephemeral inline volumes, matching the existing PV/StorageClass mountOptions behavior.

- --attr-cache-max-size-mb caps attribute-cache memory (0 = auto-tune to 1% of RAM, clamped [64 MB, 1 GB]).
- --kernel-list-cache-timeout sets the kernel directory-listing cache TTL in seconds (fuse3 only; 0 = disabled).

Extend TestSanitizeMountOptions with a case covering both flags and document them in docs/driver-parameters.md.

Copilot-Session: ea698678-2287-4533-8a91-29810a888a64

**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


andyzhangx (commented):
/ok-to-test

andyzhangx (approved):
/lgtm

PR #2601, opened 2026-07-31
fix: rewrite mountPath alongside hostPath when linux.distro=rhcos

### LINKED ISSUE #2599: linux.distro=rhcos (1.27.8) does not fix crun Init:CreateContainerError: mountPath still /host/usr/local
## Summary

The `linux.distro=rhcos` support added in chart 1.27.8 (#2585 / #2586, following #2473) does not fix the crun `Init:CreateContainerError` it was intended to fix. It rewrites the **hostPath** side of the `host-usr-local` volume to `/var/usrlocal`, but leaves the in-container **mountPath** at `/host/usr/local`. crun validates the mount destination path as well as the source, so the original failure still occurs.

Tested on ARO (Azure Red Hat OpenShift), RHCOS nodes, `defaultRuntime: crun`.

## Observed

Setting `linux.distro=rhcos` with `node.hostPath.usrLocal` left at its default on chart 1.27.8, the `install-blobfuse-proxy` init container fails with the identical error reported in #2473:

```
Warning  Failed  6s (x13 over 2m25s)  kubelet
  spec.initContainers{install-blobfuse-proxy}: Error: container create failed:
  creating `/host/usr/local`: openat2 `host/usr/local`: No such file or directory
```

Pod stays in `Init:CreateContainerError`.

## Cause

`templates/csi-blob-node.yaml` rewrites only `$usrLocalHostPath`:

```gotemplate
{{- $usrLocalHostPath := .Values.node.hostPath.usrLocal -}}
{{- if and (eq $usrLocalHostPath "/usr/local") (eq .Values.linux.distro "rhcos") -}}
{{- $usrLocalHostPath = "/var/usrlocal" -}}
{{- end }}
```

Rendered output with `linux.distro=rhcos` on 1.27.8:

```yaml
# volumeMounts
- name: host-usr-local
  mountPath: /host/usr/local     # <-- still the symlinked path
# volumes
- name: host-usr-local
  hostPath:
    path: /var/usrlocal          # <-- correctly rewritten
```

On RHCOS, `/usr/local` is a symlink to `/var/usrlocal`, and `/host/usr` is bind-mounted read-only from `/usr`. crun resolves the destination with `openat2()` + `RESOLVE_NO_SYMLINKS`, so creating the mountpoint `/host/usr/local` fails with `ENOENT`. runc tolerates this, which is why the problem is crun-specific.

The template's own comment states the mountPath is deliberately left in place *"so install scripts (default BIN_PATH=/usr/local/bin) keep working unchanged"* — but that only holds under runc.

## Workaround (works)

Setting the override explicitly, as enabled by #2366, rewrites **both** sides and works correctly under crun:

```yaml
linux:
  distro: fedora
node:
  hostPath:
    usrLocal: /var/usrlocal
```

Renders as:

```yaml
- name: host-usr-local
  mountPath: /host/var/usrlocal   # both sides avoid the symlink
- name: host-usr-local
  hostPath:
    path: /var/usrlocal
```

This is what we had before trying `distro=rhcos`, and it has been stable across a full crun migration on multiple clusters. We have reverted to it.

## Suggested fix

Rewrite the `mountPath` alongside the `hostPath` when the rhcos auto-rewrite triggers, so both sides point at the non-symlinked path (matching what the explicit override already achieves). If `BIN_PATH` needs to remain `/usr/local/bin`, that could be set independently of the mount destination.

Happy to test a patch.

## Environment

- chart 1.27.8 (also reproduced the underlying behaviour on 1.27.7 without the override)
- ARO, RHCOS nodes, `defaultRuntime: crun` via `ContainerRuntimeConfig`
- `node.enableBlobfuseProxy: false`, `blobfuseProxy.installBlobfuse: false`, `installBlobfuse2: false` — note the `install-blobfuse-proxy` init container is added unconditionally, so the mount is created even when all install flags are false

## Related

- #2473 — original ARO `Init:CreateContainerError` report
- #2585 / #2586 — the PRs that added `linux.distro=rhcos`
- #2366 / #2376 / #2377 — added the `node.hostPath.*` overrides (the working workaround)

andyzhangx:

Fixes #2599.

The `linux.distro=rhcos` auto-rewrite added in #2585 / #2586 only rewrote the **hostPath** side of the `host-usr-local` volume to `/var/usrlocal` but left the in-container **mountPath** at `/host/usr/local`. crun validates the mount destination with `openat2(RESOLVE_NO_SYMLINKS)` and refuses to create `/host/usr/local` because `/usr/local` is a symlink to `/var/usrlocal` on RHCOS (and `/host/usr` is bind-mounted read-only from `/usr`), so the `install-blobfuse-proxy` init container still fails with `Init:CreateContainerError` on ARO clusters with `defaultRuntime: crun`.

This PR rewrites **both** sides to `/var/usrlocal` when the auto-rewrite triggers, so neither the hostPath source nor the container destination traverses the symlink. This matches the known-working explicit override (`node.hostPath.usrLocal=/var/usrlocal`) that reporters have been using since #2366.

Also adds an `rhcos` branch to `install-proxy-rhcos.sh` so `BIN_PATH` points at `/var/usrlocal/bin` inside the init container. On the host, `/var/usrlocal/bin` is the same real directory as `/usr/local/bin` via the OSTree symlink, so the resulting `blobfuse-proxy.service` unit resolves the binary correctly.

**Rendered diff with `linux.distro=rhcos`**:

Before:
```yaml
- name: host-usr-local
  mountPath: /host/usr/local     # <-- crun fails here
hostPath:
  path: /var/usrlocal
```

After:
```yaml
- name: host-usr-local
  mountPath: /host/var/usrlocal  # both sides avoid the symlink
hostPath:
  path: /var/usrlocal
```

Verified with `helm template` for the default, `linux.distro=rhcos`, and explicit `node.hostPath.usrLocal=/var/usrlocal` cases.

Fixes #2599

**Requirements**:

- [x] squashed commits
- [x] includes documentation (chart template comment updated)
- [ ] adds unit tests
- [x] tested manually via `helm template` for default / rhcos / explicit-override cases

**Release note**:
```release-note
Fix `linux.distro=rhcos` still hitting crun `Init:CreateContainerError` on RHCOS/ARO nodes: rewrite the `host-usr-local` in-container mountPath to `/host/var/usrlocal` alongside the already-rewritten hostPath, and point BIN_PATH at `/var/usrlocal/bin` in `install-proxy-rhcos.sh`.
```


andyzhangx:
Addressed in 007b639: switched to `BIN_PATH=${BIN_PATH:-/var/usrlocal/bin}` so a preset `BIN_PATH` env still wins, matching the default branch's behavior. Users who need to override the layout further can still set `CUSTOM_BIN_PATH`.

PR #2604, opened 2026-08-03
fix: ensure consistent volume attribute key resolution across code paths

andyzhangx:
Add NormalizeVolumeAttributes to lowercase and deduplicate keys in the volumeAttributes map at the entry of NodePublishVolume and NodeStageVolume. This ensures all downstream consumers (useWorkloadIdentity, GetAuthEnv, ephemeral detection, etc.) resolve the same deterministic value for each logical key.

- Add unit tests for NormalizeVolumeAttributes.


andyzhangx:
Closing in favor of #2603 which includes this fix (NormalizeVolumeAttributes) along with unconditional containerName validation.

PR #2620, opened 2026-08-12
fix: harden server field handling in inline volume with managed identity auth mount

kapilupadhayay:
Harden server field handling for inline volumes.

**What type of PR is this?**
/kind bug

Validates inline volume server values against the resolved storage account
and configured cloud environment. Existing PersistentVolume and
StorageClass endpoint behavior remains unchanged.

Fixes #

**Requirements**:
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```release-note
Improve server field validation for inline Blob CSI volumes.
```

### DIFF
--- a/pkg/blob/blob.go
+++ b/pkg/blob/blob.go
@@ -1349,7 +1349,7 @@ func isAzureDNSZoneStorageHost(host, account string) bool {
 		return false
 	}
 	if len(zoneID) == 2 &&
-		(zoneID[0] < '0' || zoneID[0] > '9' ||
+		(zoneID[0] < '1' || zoneID[0] > '9' ||
 			zoneID[1] < '0' || zoneID[1] > '9') {
 		return false
 	}
--- a/pkg/blob/blob.go
+++ b/pkg/blob/blob.go
@@ -1306,6 +1306,8 @@ func validateInlineVolumeServerHost(server, host, accountName, storageEndpointSu
 	}
 
 	secondaryAccount := account + "-secondary"
+	// Secondary DFS forms are accepted for structural consistency. They do not
+	// currently resolve for HNS accounts because HNS does not support RA-GRS.
 	allowedHosts := map[string]struct{}{
 		fmt.Sprintf("%s.blob.%s", account, suffix):                      {},
 		fmt.Sprintf("%s.dfs.%s", account, suffix):                       {},
@@ -1319,7 +1321,8 @@ func validateInlineVolumeServerHost(server, host, accountName, storageEndpointSu
 	if _, ok := allowedHosts[host]; ok {
 		return nil
 	}
-	if suffix == defaultStorageEndPointSuffix && isAzureDNSZoneStorageHost(host, account) {
+	if suffix == defaultStorageEndPointSuffix &&
+		isAzureDNSZoneStorageHost(host, account) {
 		return nil
 	}
 
@@ -1330,20 +1333,41 @@ func validateInlineVolumeServerHost(server, host, accountName, storageEndpointSu
 // format without requiring a storage management-plane lookup from the node.
 func isAzureDNSZoneStorageHost(host, account string) bool {
 	parts := strings.Split(host, ".")
-	if len(parts) != 6 {
+	if len(parts) != 6 && len(parts) != 7 {
 		return false
 	}
 	if parts[0] != account && parts[0] != account+"-secondary" {
 		return false
 	}
 	zone := parts[1]
-	if len(zone) != 3 || zone[0] != 'z' || zone[1] < '0' || zone[1] > '9' || zone[2] < '0' || zone[2] > '9' {
+	zoneID, hasZonePrefix := strings.CutPrefix(zone, "z")
+	if !hasZonePrefix {
 		return false
 	}
-	if parts[2] != "blob" && parts[2] != "dfs" {
+	if len(zoneID) == 1 &&
+		(zoneID[0] < '1' || zoneID[0] > '9') {
 		return false
 	}
-	return parts[3] == "storage" && parts[4] == "azure" && parts[5] == "net"
+	if len(zoneID) == 2 &&
+		(zoneID[0] < '0' || zoneID[0] > '9' ||
+			zoneID[1] < '0' || zoneID[1] > '9') {
+		return false
+	}
+	if len(zoneID) < 1 || len(zoneID) > 2 {
+		return false
+	}
+
+	serviceIndex := 2
+	if len(parts) == 7 {
+		if parts[serviceIndex] != "privatelink" {
+			return false
+		}
+		serviceIndex++
+	}
+	if parts[serviceIndex] != "blob" && parts[serviceIndex] != "dfs" {
+		return false
+	}
+	return strings.Join(parts[serviceIndex+1:], ".") == "storage.azure.net"
 }
 
 // parseInlineVolumeServer extracts a normalized hostname from an HTTPS server
--- a/pkg/blob/blob.go
+++ b/pkg/blob/blob.go
@@ -22,6 +22,8 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"net"
+	"net/url"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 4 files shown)


pkg/blob/nodeserver.go
```diff
@@ -404,6 +405,16 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 		return nil, status.Errorf(codes.InvalidArgument, "NodeStageVolume: containerName must be specified for ephemeral volumes")
 	}
 
+	trustedStorageEndpointSuffix := d.getStorageEndPointSuffix()
+	if ephemeralVol {
```
andyzhangx: since this is only for inline + MSI authentication, why not add such restriction in L409?
```
if ephemeralVol && authType is MSI) {
    validate server ...
}
```

kapilupadhayay: Fixed, thanks

pkg/blob/nodeserver.go
```diff
@@ -440,13 +453,30 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
...
+		originalServerAddress := serverAddress
+		serverAddress, err = parseInlineVolumeServer(originalServerAddress)
+		if err != nil {
+			return nil, status.Errorf(codes.InvalidArgument, "NodeStageVolume: %v", err)
+		}
+		err = validateInlineVolumeServerHost(originalServerAddress, serverAddress, accountName, trustedStorageEndpointSuffix)
+		if err != nil {
+			accountEndpoints, endpointErr := d.getStorageAccountEndpoints(ctx, getValueInMap(attrib, subscriptionIDField), resourceGroup, accountName)
+			if endpointErr == nil {
+				err = validateInlineVolumeServerHost(originalServerAddress, serverAddress, accountName, trustedStorageEndpointSuffix, accountEndpoints...)
```
andyzhangx: what's the intension of the second validateInlineVolumeServerHost call here?

pkg/blob/nodeserver.go
```diff
@@ -440,13 +453,30 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
...
 	}
+	if isInlineMSI {
+		originalServerAddress := serverAddress
+		serverAddress, err = parseInlineVolumeServer(originalServerAddress)
+		if err != nil {
+			return nil, status.Errorf(codes.InvalidArgument, "NodeStageVolume: %v", err)
+		}
+		err = validateInlineVolumeServerHost(originalServerAddress, serverAddress, accountName, trustedStorageEndpointSuffix)
+		if err != nil {
+			accountEndpoints, endpointErr := d.getStorageAccountEndpoints(ctx, getValueInMap(attrib, subscriptionIDField), resourceGroup, accountName)
```
andyzhangx: note that `getStorageAccountEndpoints` call may fail here since by default, the node daemonset does not have permission to get a storage account properties by default.

pkg/blob/blob.go
```diff
@@ -1281,6 +1283,98 @@ func ValidateContainerName(containerName string) error {
...
 }
 
+// ValidateInlineVolumeServer ensures a namespace-controlled inline volume can
+// only direct storage traffic to the resolved account in the configured cloud.
+func ValidateInlineVolumeServer(server, accountName, storageEndpointSuffix string) error {
+	_, err := normalizeInlineVolumeServer(server, accountName, storageEndpointSuffix)
+	return err
+}
+
+func normalizeInlineVolumeServer(server, accountName, storageEndpointSuffix string, accountEndpoints ...string) (string, error) {
```
andyzhangx: pls add comments for each new func

pkg/blob/blob.go
```diff
@@ -1281,6 +1283,98 @@ func ValidateContainerName(containerName string) error {
...
+	if err != nil {
+		return "", err
+	}
+	if err := validateInlineVolumeServerHost(server, host, accountName, storageEndpointSuffix, accountEndpoints...); err != nil {
+		return "", err
+	}
+	return host, nil
+}
+
+func validateInlineVolumeServerHost(server, host, accountName, storageEndpointSuffix string, accountEndpoints ...string) error {
```
andyzhangx: pls add comments for each new func

pkg/blob/blob.go
```diff
@@ -1281,6 +1283,98 @@ func ValidateContainerName(containerName string) error {
...
+		}
+	}
+	if _, ok := allowedHosts[host]; !ok {
+		return fmt.Errorf("invalid server %q: hostname must match storage account %q in the configured Azure cloud", server, accountName)
+	}
+
+	return nil
+}
+
+func parseInlineVolumeServer(server string) (string, error) {
```
andyzhangx, line 1375: pls add comments for each new func

pkg/blob/nodeserver.go
```diff
@@ -645,6 +676,45 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
...
+		subscriptionID = d.cloud.SubscriptionID
+	}
+	if resourceGroup == "" {
+		resourceGroup = d.cloud.ResourceGroup
+	}
+	accountClient, err := d.cloud.ComputeClientFactory.GetAccountClientForSub(subscriptionID)
+	if err != nil {
+		return nil, err
+	}
+	account, err := accountClient.GetProperties(ctx, resourceGroup, accountName, nil)
```
landreasyan: Could this result in throttling and should we consider having a cache here?

pkg/blob/nodeserver.go
```diff
@@ -645,6 +676,45 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
...
+	if err != nil {
+		return nil, err
```
landreasyan: For consistency, can we wrap this error? 
`status.Errorf(codes.Internal, "NodeStageVolume: failed to get storage account endpoints: %v", endpointErr)`

pkg/blob/nodeserver.go
kapilupadhayay: That was not right, thanks. Replaced with one normalization-and-validation call

pkg/blob/nodeserver.go
kapilupadhayay: removed

pkg/blob/blob.go
kapilupadhayay: added

pkg/blob/blob.go
kapilupadhayay: added

pkg/blob/blob.go
kapilupadhayay: added

pkg/blob/nodeserver.go
kapilupadhayay: code removed

pkg/blob/nodeserver.go
kapilupadhayay: code removed

pkg/blob/nodeserver.go
```diff
@@ -440,13 +451,30 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
...
+		serverAddress, err = parseInlineVolumeServer(originalServerAddress)
+		if err != nil {
+			return nil, status.Errorf(codes.InvalidArgument, "NodeStageVolume: %v", err)
+		}
+		err = validateInlineVolumeServerHost(originalServerAddress, serverAddress, accountName, trustedStorageEndpointSuffix)
+		if err != nil {
+			accountEndpoints, endpointErr := d.getStorageAccountEndpoints(ctx, getValueInMap(attrib, subscriptionIDField), resourceGroup, accountName)
+			if endpointErr == nil {
+				err = validateInlineVolumeServerHost(originalServerAddress, serverAddress, accountName, trustedStorageEndpointSuffix, accountEndpoints...)
+			}
```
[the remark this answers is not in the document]
kapilupadhayay: The ARM endpoint lookup has been removed. Inline MSI endpoints are now validated deterministically, avoiding node identity permissions, transient ARM failures, and throttling. Invalid endpoints still return  InvalidArgument .

andyzhangx:
A few endpoint formats worth double-checking against the validator:

**Likely missing:**

1. **Privatelink + DNS-zone endpoints combined**
   `<account>[-secondary].z[00-99].privatelink.(blob|dfs).storage.azure.net`
   Storage accounts using the newer Azure DNS zone endpoints behind a Private Endpoint hit this shape. If the regex accepts `privatelink.(blob|dfs).<suffix>` OR `z##.(blob|dfs).storage.azure.net` but not the combined form, valid customer configs will be rejected.

2. **DNS-zone endpoints in sovereign clouds**
   The `z00–z99` case is hardcoded to `storage.azure.net`, which is public-cloud only. Azure DNS zone endpoints are also available in Mooncake / USGov with different suffixes. If this fix is expected to work in sovereign clouds, the suffix here should be parametrized rather than a hard `storage.azure.net`.

**Minor note:**

- `-secondary` on `dfs`: RA-GRS secondary endpoints don't actually exist for HNS accounts (HNS doesn't support GRS/RA-GRS), so `<account>-secondary.dfs.<suffix>` won't resolve in practice. Not a bug — just worth noting in the docstring so it doesn't look like an intended path.

**Correctly excluded** (not missing): custom domains, ADLS Gen1 (`azuredatalakestore.net`), and non-blob services (`file`/`queue`/`table`).

Recommendation: at minimum add case (1) — it's a real, customer-hittable format on newer storage accounts with Private Endpoints.

kapilupadhayay:
Thanks @andyzhangx . I added support for the combined Private Link and Azure DNS-zone
endpoint forms:
<account>[-secondary].z<N>.privatelink.blob.storage.azure.net
<account>[-secondary].z<N>.privatelink.dfs.storage.azure.net
Unit tests cover primary and secondary Blob and DFS combinations, including
the complete  NodeStageVolume  path.

I also documented that secondary DFS forms are accepted for structural
consistency, although they do not currently resolve for HNS accounts because
HNS does not support RA-GRS.

I have not added sovereign-cloud DNS-zone formats. Microsoft currently
documents this preview only for public Azure regions, so we can revisit that
separately if authoritative sovereign endpoint formats become available.

andyzhangx:
One minor consistency point on `isAzureDNSZoneStorageHost` (non-blocking, feel free to defer to a follow-up):

The current zone-ID validation accepts `z1..z9` and `z00..z99` but rejects `z0`. In practice Azure DNS-zone IDs are `z1..z99` with no leading-zero padding, so `z0`, `z00`, and `z01` all shouldn't exist in the wild. The current shape is a bit in-between — it rejects single-digit `z0` but happily accepts two-digit `z00`/`z01`.

Two consistent options:

- **Strict (match reality):** only allow `z1..z99` — reject any leading zero. Concretely, add `zoneID[0] != '0'` to the `len(zoneID) == 2` branch, which also naturally rejects `z0` for free.
- **Loose (pure structural):** accept anything that parses as a 1–2 digit number, including `z0`/`z00`/`z01`.

I'd lean toward the strict version — it matches what Azure actually hands out and keeps the semantics readable as "`z1..z99`". Not a blocker for this PR.

andyzhangx (approved):
/lgtm

PR #2630, opened 2026-08-18
fix: protect storageAccountName for inline volumes

kapilupadhayay:
**What type of PR is this?**
/kind bug

Clears the compatibility `storageAccountName` field together with `storageAccount` when inline volume key access with the driver identity is disabled. This prevents the alias from bypassing the existing inline-volume protection.

Persistent volumes, workload identity mounts, and clusters that explicitly enable inline volume key access with identity remain unchanged.

Fixes #

**Requirements**:
- [ ] includes documentation
- [x] adds unit tests

**Validation**:
- Complete `pkg/blob` unit test suite passes.
- Protected inline volumes clear both account-name fields.
- Persistent volumes preserve `storageAccountName`.
- Workload identity and explicit administrator opt-in paths preserve `storageAccountName`.

**Release note**:
```release-note
Protect the storageAccountName compatibility field for inline volumes.
```


andyzhangx (commented):
/ok-to-test

andyzhangx (approved):
/lgtm

PR #2635, opened 2026-08-19
fix: block preload for inline volumes

kapilupadhayay:
**What type of PR is this?**
/kind bug

Rejects the BlobFuse2 `--preload` option for inline volumes. Preloading can download the entire container into host storage outside the requesting Pod's ephemeral-storage accounting, allowing an untrusted Pod creator to consume node storage.

The driver returns `InvalidArgument` with guidance to use a StorageClass-based persistent volume instead. Administrator-controlled persistent-volume mount options remain unchanged.

Fixes #

**Requirements**:
- [ ] includes documentation
- [x] adds unit tests

**Validation**:
- Complete `pkg/blob` unit test suite passes.
- Inline `--preload` and `--preload=true` are rejected.
- Persistent-volume `--preload` remains supported through CSI mount flags.

**Release note**:
```release-note
Reject the BlobFuse2 preload option for inline volumes and direct users to StorageClass-based persistent volumes.
```

### DIFF
--- a/pkg/blob/blob.go
+++ b/pkg/blob/blob.go
@@ -268,7 +268,6 @@ var (
 		"--no-symlinks":                    {},
 		"--pool-size":                      {},
 		"--pre-mount-validate":             {},
-		"--preload":                        {},
 		"--preserve-acl":                   {},
 		"--read-only":                      {},
 		"--required-free-space-mb":         {},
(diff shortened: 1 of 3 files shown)


andyzhangx (commented):
/ok-to-test

andyzhangx (approved):
/lgtm

PR #2679, opened 2026-09-09
fix: reject non-ASCII volume attribute keys on inline volumes

kapilupadhayay:
**What type of PR is this?**

/kind bug

Volume attribute key validation uses `strings.ToLower` for duplicate detection, while `getValueInMap`/`setKeyValueInMap` look up and mutate keys with `strings.EqualFold`. The two disagree on non-ASCII input: `ToLower` leaves U+017F (LATIN SMALL LETTER LONG S) unchanged, so `ſtorageAccount` does not collide with `storageaccount`, but `EqualFold` reports the two as equal.

This matters for inline (ephemeral) volumes, where the attributes are pod-author controlled. `NodePublishVolume` blanks `storageAccount`/`storageAccountName` so the account key must come from a Secret. `setKeyValueInMap` only rewrites the *first* `EqualFold` match, and Go map iteration order is randomized, so a Unicode decoy key can absorb that blanking and leave the real attacker-supplied account name in place.

This PR rejects any volume attribute key containing a non-ASCII byte, before any lookup or mutation happens. The check is scoped to inline volumes:

- `ValidateASCIIVolumeAttributeKeys` is new and runs only when `csi.storage.k8s.io/ephemeral` is set.
- The existing `ValidateVolumeAttributeKeys` case-collision check is unchanged and continues to run for **every** volume.
- Persistent volume and StorageClass behaviour is therefore unchanged, including for non-ASCII keys.

**Requirements**:
- [ ] includes documentation
- [x] adds unit tests

Deliberately scoped to inline volumes to avoid a behaviour change for existing persistent volumes. An earlier iteration applied the ASCII check globally and also lowercased accepted keys in place, which changed behaviour for `StorageClass`-provisioned volumes; that was reverted.

Verified on an AKS cluster (v1.35.7) with the driver built from this branch, A/B against a build of `master`:

- Inline volume carrying both `storageaccount` and the `ſtorageaccount` decoy — on `master` the pod reaches `Running`; with this change the mount is refused:
  ```
  MountVolume.SetUp failed for volume "blobvol" : rpc error: code = InvalidArgument
  desc = NodePublishVolume: invalid volume attribute key "ſtorageaccount": only ASCII characters are allowed
  ```
- Inline volume with ordinary ASCII keys still mounts and is read-write.
- Persistent volume carrying a non-ASCII attribute key still mounts and is read-write, confirming the new check does not apply there.

**Release note**:
```
none
```

### DIFF
--- a/pkg/blob/blob.go
+++ b/pkg/blob/blob.go
@@ -32,6 +32,7 @@ import (
 	"sync"
 	"time"
 	"unicode"
+	"unicode/utf8"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/storage/armstorage"
 	azstorage "github.com/Azure/azure-sdk-for-go/storage"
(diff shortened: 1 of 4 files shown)


andyzhangx (commented):
/ok-to-test

kapilupadhayay:
Thanks @copilot-pull-request-reviewer — good catch, the `NodeStageVolume` guard genuinely had no handler-level coverage.

Added in 3c3b0975b:

- `TestNodeStageVolume/[Error] ephemeral volume rejects non-ASCII attribute keys` — ephemeral context with `ſtorageAccount`, asserts `InvalidArgument` from `NodeStageVolume` directly.
- `TestNodeStageVolume/persistent volume keeps existing behavior for non-ASCII attribute keys` — same non-ASCII key without `csi.storage.k8s.io/ephemeral`, made to fail later on `mountPermissions` instead. Because it fails with a *different* error, it proves the ASCII guard did not fire and the scoping to inline volumes holds.

`go test ./pkg/blob -count=1` passes, `gofmt`/`go vet` clean.

andyzhangx (approved):
/lgtm

PR #2682, opened 2026-09-13
Adds BlobFuse2 distributed-cache support and default DNS server option to the Blob CSI driver.

MahashriSaini:
Adds support for blobfuse2 distributed cache.
1. Allows the following distributed-cache mount options:
  - `--distributed-cache-discovery-endpoint`
  - `--distributed-cache-dns-server`
  - `--distributed-cache-node-ttl`
  - `--distributed-cache-block-size`
  - `--distributed-cache-node-memory`
  - `--distributed-cache-prefetch`
  - `--distributed-cache-parallelism`
 2.  Adds `--distributed-cache-dns-server=10.0.0.10` when distributed cache is enabled and the user does not provide a DNS server.

Validation:
1. Unit Tests and Integration Tests
2. Local Manual E2E testing on AKS cluster
- Confirmed the generated BlobFuse2 command includes the default DNS server
- Confirmed a user-provided DNS server overrides the default.
- Confirmed distributed-cache L2 lookup, miss handling, Azure download, and asynchronous cache upload.

The BlobFuse2 package version is not changed in this PR. It can be updated separately after a BlobFuse2 release containing distributed-cache support is available.

**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/blob/blob.go
```diff
@@ -1594,14 +1604,31 @@ func appendDefaultMountOptions(mountOptions []string, tmpPath, containerName str
...
 		for k := range defaultMountOptions {
 			if strings.HasPrefix(mountOption, k) {
 				included[k] = true
 			}
 		}
 	}
+	if distributedCacheEnabled {
+		included["--tmp-path"] = true
+		included["--empty-dir-check"] = true
+		defaultMountOptions["--distributed-cache-dns-server"] = "10.0.0.10"
```
kapilupadhayay: Lets make a constant for this value set on the top and then use that constant

kapilupadhayay: Please add a comment to specify the intent why this value for DNS is being used as a default for a constant

kapilupadhayay (changes requested):

pkg/blob/blob.go
MahashriSaini: Done

andyzhangx:
/hold

until  BlobFuse2 release containing distributed-cache support is available

pkg/blob/blob.go
```diff
@@ -1535,8 +1535,7 @@ func SanitizeMountOptions(mountOptions []string) ([]string, error) {
 		if _, ok := allowedEphemeralMountOptions[flagName]; !ok {
 			return nil, fmt.Errorf("mount option %q is not allowed for ephemeral volumes", flagName)
 		}
-		if (flagName == "--block-cache-parallelism" ||
-			flagName == "--distributed-cache-parallelism") && len(parts) != 2 {
+		if flagName == "--block-cache-parallelism" && len(parts) != 2 {
```
kapilupadhayay: The gate accepts bare --distributed-cache-parallelism and --distributed-cache-parallelism=. During the compatibility window, fallback removes the malformed option and the mount succeeds. After BlobFuse gains Tachyon support, the same option is forwarded to its typed uint32 flag parser and the mount fails.

**Required Change:** 
Required change: continue requiring a nonempty integer from  1  to  128  for  --distributed-cache-parallelism , regardless of whether the current BlobFuse supports Tachyon. Capability fallback should process only syntactically valid distributed-cache requests.
The required code behavior is essentially:

```
if flagName == "--distributed-cache-parallelism" {
    if len(parts) != 2 || parts[1] == "" {
        return nil, fmt.Errorf(
            "mount option %s requires a value",
            flagName,
        )
    }

    parallelism, err := strconv.ParseUint(parts[1], 10, 32)
    if err != nil || parallelism == 0 || parallelism > 128 {
        return nil, fmt.Errorf(
            "mount option %s must be an integer between 1 and 128",
            flagName,
        )
    }
}
```

kapilupadhayay (approved):

pkg/blob/blob.go
MahashriSaini: Thank you Kapil for detailed review and suggestion. Done

pkg/blobfuse-proxy/pb/azure_blob_mount_grpc.pb.go
```diff
@@ -84,17 +97,36 @@ func _MountService_MountAzureBlob_Handler(srv interface{}, ctx context.Context,
...
+		Server:     srv,
+		FullMethod: "/MountService/GetBlobfuseCapabilities",
+	}
+	handler := func(ctx context.Context, req interface{}) (interface{}, error) {
+		return srv.(MountServiceServer).GetBlobfuseCapabilities(ctx, req.(*BlobfuseCapabilitiesRequest))
+	}
+	return interceptor(ctx, in, info, handler)
+}
+
+var _MountService_serviceDesc = grpc.ServiceDesc{
```
croomes: Regenerating this file removes the exported `MountService_ServiceDesc` symbol and replaces it with the private `_MountService_serviceDesc`. Since `pkg/blobfuse-proxy/pb` is importable, downstream code that registers or inspects the exported descriptor will no longer compile. Is there a reason to change this? If not, please preserve or create an exported compatibility alias.

MahashriSaini: It was caused by using a different version of  protoc-gen-go-grpc . I will use the pinned version and correct it.

croomes (approved):
LGTM - thanks

andyzhangx (commented):
/hold cancel
/lgtm

andyzhangx (approved):

